PR #15, opened 2018-01-16
Implement serialization

tuxxy:
### What this does:
1. Implements the following methods on `KFrag`, `CapsuleFrag`, `Capsule`, `ReconstructedCapsule`, and `ChallengeResponse`.
    - `from_bytes`
    - `to_bytes`
    - `__bytes__`
2. Adds tests for the serialization and deserialization functions in `test_umbral`.

### DIFF
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -101,6 +101,24 @@ def test_cfrag_serialization():
     assert new_cfrag.point_eph_ni == cfrag.point_eph_ni
 
 
+def test_capsule_serialization():
+    pre = umbral.PRE(umbral.UmbralParameters())
+
+    priv_key = pre.gen_priv()
+    pub_key = pre.priv2pub(priv_key)
+
+    _, capsule = pre.encapsulate(pub_key)
+    capsule_bytes = capsule.to_bytes()
+
+    assert len(capsule_bytes) == 98
+
+    new_capsule = umbral.Capsule.from_bytes(capsule_bytes,
+                                            umbral.UmbralParameters().curve)
+    assert new_capsule.point_eph_e == capsule.point_eph_e
+    assert new_capsule.point_eph_v == capsule.point_eph_v
+    assert new_capsule.bn_sig == capsule.bn_sig
+
+
 # @pytest.mark.parametrize("N,threshold", parameters)
 # def test_cheating_Ursula_replays_old_reencryption(N, threshold):
 #     pre = umbral.PRE()
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -144,6 +144,34 @@ def test_reconstructed_capsule_serialization():
     assert new_rec_capsule.v_prime == rec_capsule.v_prime
     assert new_rec_capsule.point_eph_ni == rec_capsule.point_eph_ni
 
+
+def test_challenge_response_serialization():
+    pre = umbral.PRE(umbral.UmbralParameters())
+
+    priv_key = pre.gen_priv()
+    pub_key = pre.priv2pub(priv_key)
+
+    _, capsule = pre.encapsulate(pub_key)
+    kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
+
+    cfrag = pre.reencrypt(kfrags[0], capsule)
+
+    capsule.attach_cfrag(cfrag)
+    ch_resp = pre.challenge(kfrags[0], capsule, cfrag)
+
+    ch_resp_bytes = ch_resp.to_bytes()
+    assert len(ch_resp_bytes) == 228
+
+    new_ch_resp = umbral.ChallengeResponse.from_bytes(
+                            ch_resp_bytes, umbral.UmbralParameters().curve)
+    assert new_ch_resp.e2 == ch_resp.e2
+    assert new_ch_resp.v2 == ch_resp.v2
+    assert new_ch_resp.point_kfrag_commitment == ch_resp.point_kfrag_commitment
+    assert new_ch_resp.point_kfrag_pok == ch_resp.point_kfrag_pok
+    assert new_ch_resp.bn_kfrag_sig1 == ch_resp.bn_kfrag_sig1
+    assert new_ch_resp.bn_kfrag_sig2 == ch_resp.bn_kfrag_sig2
+    assert new_ch_resp.bn_sig == ch_resp.bn_sig
+
 # @pytest.mark.parametrize("N,threshold", parameters)
 # def test_cheating_Ursula_replays_old_reencryption(N, threshold):
 #     pre = umbral.PRE()
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -72,7 +72,7 @@ def test_kfrag_serialization():
     new_frag = umbral.KFrag.from_bytes(frag_bytes,
                                        umbral.UmbralParameters().curve)
     assert new_frag.bn_id == frags[0].bn_id
-    assert new_frag.point_key == frags[0].point_key
+    assert new_frag.bn_key == frags[0].bn_key
     assert new_frag.point_eph_ni == frags[0].point_eph_ni
     assert new_frag.point_commitment == frags[0].point_commitment
     assert new_frag.bn_sig1 == frags[0].bn_sig1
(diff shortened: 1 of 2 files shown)


tests/test_umbral.py
```diff
@@ -58,6 +58,120 @@ def test_m_of_n(N, threshold):
     assert sym_key_2 == sym_key
 
 
+def test_kfrag_serialization():
+    pre = umbral.PRE(umbral.UmbralParameters())
```
jMyles: groovy.

tests/test_umbral.py
```diff
@@ -58,6 +58,120 @@ def test_m_of_n(N, threshold):
...
+    assert new_frag.bn_sig2 == frags[0].bn_sig2
+
+
+def test_cfrag_serialization():
+    pre = umbral.PRE(umbral.UmbralParameters())
+
+    priv_key = pre.gen_priv()
+    pub_key = pre.priv2pub(priv_key)
+
+    _, capsule = pre.encapsulate(pub_key)
```
jMyles: that's a beaut.

tests/test_umbral.py
```diff
@@ -58,6 +58,120 @@ def test_m_of_n(N, threshold):
...
+    kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
+
+    cfrag = pre.reencrypt(kfrags[0], capsule)
+    cfrag_bytes = cfrag.to_bytes()
+
+    assert len(cfrag_bytes) == 131
+
+    new_cfrag = umbral.CapsuleFrag.from_bytes(cfrag_bytes,
```
jMyles: Great naming, @tuxxy - your tests have definitely become more readable lately.  Keep it up!

tests/test_umbral.py
```diff
@@ -58,6 +58,120 @@ def test_m_of_n(N, threshold):
...
+def test_capsule_serialization():
+    pre = umbral.PRE(umbral.UmbralParameters())
+
+    priv_key = pre.gen_priv()
+    pub_key = pre.priv2pub(priv_key)
+
+    _, capsule = pre.encapsulate(pub_key)
+    capsule_bytes = capsule.to_bytes()
+
+    assert len(capsule_bytes) == 98
```
jMyles: The assertion is great, but I think the reader will be wanting to know *why* this is supposed to be 98 bytes (or above, 131 for the cFrag, 99 for  rec_capsule_bytes, and so on down the module).

tests/test_umbral.py
```diff
@@ -58,6 +58,120 @@ def test_m_of_n(N, threshold):
...
+
+    new_capsule = umbral.Capsule.from_bytes(capsule_bytes,
```
jMyles: Wow, what an upgrade.  This reads like a poem.

tests/test_umbral.py
```diff
@@ -58,6 +58,120 @@ def test_m_of_n(N, threshold):
...
+    pub_key = pre.priv2pub(priv_key)
+
+    _, capsule = pre.encapsulate(pub_key)
+    kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
+
+    cfrag = pre.reencrypt(kfrags[0], capsule)
+
+    capsule.attach_cfrag(cfrag)
+
+    rec_capsule = capsule.reconstruct()
```
jMyles: So, here, I think I prefer a mutation rather than an instantiation.  It gets a little dizzying to have two objects after we refactored them to share a life cycle.  But that can come in a later PR.

umbral/umbral.py
```diff
@@ -22,6 +22,33 @@ def __init__(self, id_, key, x, u1, z1, z2):
...
+        Instantiate a KFrag object from the serialized data.
+        """
+        id = BigNum.from_bytes(data[0:32], curve)
+        key = BigNum.from_bytes(data[32:64], curve)
+        eph_ni = Point.from_bytes(data[64:97], curve)
+        commitment = Point.from_bytes(data[97:130], curve)
+        sig1 = BigNum.from_bytes(data[130:162], curve)
+        sig2 = BigNum.from_bytes(data[162:194], curve)
+
+        return KFrag(id, key, eph_ni, commitment, sig1, sig2)
```
jMyles: Why take `cls` as an argument and then not use it in this line?

umbral/umbral.py
```diff
@@ -22,6 +22,33 @@ def __init__(self, id_, key, x, u1, z1, z2):
...
+        Serialize the KFrag into a bytestring.
+        """
+        id = self.bn_id.to_bytes()
+        key = self.point_key.to_bytes()
+        eph_ni = self.point_eph_ni.to_bytes()
+        commitment = self.point_commitment.to_bytes()
+        sig1 = self.bn_sig1.to_bytes()
+        sig2 = self.bn_sig2.to_bytes()
+
+        return id + key + eph_ni + commitment + sig1 + sig2
```
jMyles: This method is a little rough I think - reading it, I have no idea why these particular attrs represent a canonical bytes repr.  At minimum, I think it's worth an additional comment.

umbral/umbral.py
```diff
@@ -49,13 +76,44 @@ def is_consistent(self, vKeys, params: UmbralParameters):
 
         return lh_exp == rh_exp
 
+    def __bytes__(self):
+        return self.to_bytes()
```
jMyles: I'll take it.

umbral/umbral.py
```diff
@@ -49,13 +76,44 @@ def is_consistent(self, vKeys, params: UmbralParameters):
...
 class CapsuleFrag(object):
     def __init__(self, e1, v1, id_, x):
         self.e1 = e1
         self.v1 = v1
         self.bn_kfrag_id = id_
         self.point_eph_ni = x
 
+    @classmethod
+    def from_bytes(cls, data: bytes, curve):
+        """
```
jMyles: I am slightly allergic to `data` as a var name.  What is this actually - cfrag_as_bytes, right?  That seems like a better name.

And then also, I don't love all the int literals as splitter fodder.  This is one of those times where I think we're better off using `ByteStringSplitter` as an separate project.  

For now, the ints are OK, but they trigger my sense of danger a little bit.

umbral/umbral.py
```diff
@@ -64,6 +122,27 @@ def __init__(self, point_eph_e, point_eph_v, bn_sig):
 
         self.cfrags = {}
 
+    @classmethod
+    def from_bytes(self, data: bytes, curve):
```
jMyles: I imagine you meant `cls` instead of `self` here, yeah?

umbral/umbral.py
```diff
@@ -64,6 +122,27 @@ def __init__(self, point_eph_e, point_eph_v, bn_sig):
...
+        """
+        Instantiates a Capsule object from the serialized data.
+        """
+        eph_e = Point.from_bytes(data[0:33], curve)
+        eph_v = Point.from_bytes(data[33:66], curve)
+        sig = BigNum.from_bytes(data[66:98], curve)
+
+        return Capsule(eph_e, eph_v, sig)
```
jMyles: Again, if this is truly a classmethod, logically speaking, let's use the class object instead of the module-level namespace here.

Or, if you want a staticmethod instead, just do that.

umbral/umbral.py
```diff
@@ -95,13 +174,40 @@ def reconstruct(self):
...
+        Instantiate ReconstructedCapsule from serialized data.
+        """
+        e_prime = Point.from_bytes(data[0:33], curve)
+        v_prime = Point.from_bytes(data[33:66], curve)
+        eph_ni = Point.from_bytes(data[66:99], curve)
+
+        return ReconstructedCapsule(e_prime, v_prime, eph_ni)
+
+    def to_bytes(self):
+        """
```
jMyles: Again, I'm left wondering *why* these attrs represent a complete repr.  If I wanted to understand the essence of bytes representation of this class, I'd look here.  So let's add a bit of docstring explaining it.

umbral/umbral.py
```diff
@@ -113,6 +219,40 @@ def __init__(self, e2, v2, u1, u2, z1, z2, z3):
         self.bn_kfrag_sig2 = z2
         self.bn_sig = z3
 
+    @classmethod
+    def from_bytes(self, data: bytes, curve):
```
jMyles: Why no type hint for curve?

umbral/umbral.py
michwill: You mean, using `cls(eph_e, ...)` if we use `classmethod`?

jMyles: Yeah, exactly.

PR #23, opened 2018-01-18
Add Key classes and add some fixes

tuxxy:
### What this does:
1. Adds the key classes `UmbralPrivateKey` and `UmbralPublicKey`.
2. Implements Scrypt-Salsa20-Poly1305 encryption/decryption for the private key, optionally.
3. Adds tests for `UmbralPrivateKey` and `UmbralPublicKey`.
4. Implements `__rmul__` on `Point` for proper scalar multiplication.
5. Fixes the notation for all scalar multiplication operations.

### Dependency additions:
1. PyNaCl -- (https://pynacl.readthedocs.io/en/stable/)

### DIFF
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -5,6 +5,7 @@ from umbral.bignum import BigNum
 from umbral.point import Point
 from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf
 
+
 class UmbralParameters(object):
     def __init__(self):
         self.curve = ec.SECP256K1()
@@ -12,6 +13,7 @@ class UmbralParameters(object):
         self.order = Point.get_order_from_curve(self.curve)
         self.h = Point.gen_rand(self.curve)
         self.u = Point.gen_rand(self.curve)
+
 
 class KFrag(object):
     def __init__(self, id_, key, x, u1, z1, z2):
@@ -56,7 +58,7 @@ class KFrag(object):
         z2 = self.bn_sig2
         x  = self.point_eph_ni
 
-        g_y = (params.g * z2) + (pub_a * z1)
+        g_y = (z2 * params.g) + (z1 * pub_a)
 
         return z1 == hash_to_bn([g_y, self.bn_id, pub_a, pub_b, u1, x], params)
     
@@ -66,12 +68,12 @@ class KFrag(object):
 
         # TODO: change this!
         h = params.h
-        lh_exp = h * self.bn_key
+        lh_exp = self.bn_key * h
 
         rh_exp = vKeys[0]
         i_j = self.bn_id
         for vKey in vKeys[1:]:
-            rh_exp = rh_exp + (vKey * i_j)
+            rh_exp = rh_exp + (i_j * vKey)
             i_j = i_j * self.bn_id
 
         return lh_exp == rh_exp
@@ -150,7 +152,7 @@ class Capsule(object):
         s = self.bn_sig
         h = hash_to_bn([e, v], params)
 
-        return params.g * s == v + (e * h)
+        return s * params.g == v + (h * e)
 
     def attach_cfrag(self, cfrag: CapsuleFrag):
         self.cfrags[cfrag.bn_kfrag_id] = cfrag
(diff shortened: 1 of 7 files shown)


umbral/umbral.py
```diff
@@ -358,13 +358,13 @@ def check_challenge(self, capsule, cFrag, challenge_resp, pub_a, pub_b):
         z2 = challenge_resp.bn_kfrag_sig2
         z3 = challenge_resp.bn_sig
 
-        g_y = (g * z2) + (pub_a * z1)
```
jMyles: So it looks like you've set `__rmul__` to `__mul__` and then reversed the order of a lot of these operations.  Is this just so that the order of operations matches the language of the paper?  Or is there a deeper / darker purpose?  :-)

tuxxy:
So, if you read the commit message, you'll see something about vector
notation.

Essentially, when performing scalar multiplication, this means putting the
scalar value to the left of the vector (or the point, in our case). So
that's what's happening there. It enhances the readability of it quite
heavily.

umbral/umbral.py
cygnusv: There are two possible *standard* conventions to represent operations between elements of algebraic groups used in cryptography:
- Multiplicative notation: This notation assumes that the operation between elements of the group is the multiplication (e.g., `g * h`), and therefore, the operation between a scalar and an element of the group is the exponentiation (in code: `g ** n`), which mathematically is equivalent to `g * g * (n times) ...`. This is the notation used when using "generic groups", which means that the scheme is not using concrete information about the underlying mathematical structure of the group (e.g., if using elliptic curves, the scheme doesn't use things like the coordinates). This notation is used in the paper, and in practically all PRE schemes out there. 
- Additive notation: It assumes that the operation is the addition (e.g., `g + h`), and therefore, the operation between a scalar and an element of the group is the multiplication (in code: `n * g`), which mathematically is equivalent to `g + g + (n times) ... `. This notation is usually used in Elliptic Curve crypto, for historic reasons. Note that the scalar multiplication is usually represented with the scalar in the left when the group is commutative. 

The problem was that the operations of the Point class followed the additive notation (which is the tradition in ECC and in libraries like OpenSSL) but our scalar multiplication had the scalar in the right. tuxxy's changes simply accommodated the code to the additive notation.

tuxxy:
Am I wrong to call this vector notation, as well or am I confusing two
standards with some commonalities?

cygnusv:
I think probably "vector notation" refers to [this](https://en.wikipedia.org/wiki/Vector_notation). Maybe the best is to refer to "additive/multiplicative notation".

tuxxy:
Alright, thanks for the insight.

tuxxy:
This is mostly good to go.

However, @cygnusv, do we want to allow these keys to be passed into the `PRE` object methods? I'm leaning towards no and, perhaps, losing the `bn_key` and `point_key` naming on the key classes and having them all be just plain `key`. Then I would expect the user to pass in `my_key.key` whenever they need to use the PRE library.

I look at these new classes as a method to handle serialization and a common place to have a bit of abstraction over `BigNum` and `Point`.

@jMyles Let me know what you think about that. I just don't want to have to handle the logic in the PRE methods using the `BigNum` and `Point` objects from within the key classes. I'd rather have people use the key classes as a means to store or transport the keys, or an interface to inherit from or something.

jMyles:
@tux, yes I tend to agree for sure.

In fact, I actually don't really like the `Capsule`, `CapsuleFrag`, and `KFrag` classes being used inside the PRE functions.  For my eyes, this is what implicates the sorts of concerns that @cygnusv has shared in #26.

I prefer that the PRE functions return tuples or dicts of elements, and that the callers then instantiate the objects.

IE, looking at `encapsulate`, instead of:

caller:
`sym_key, capsule = pre.encapsulate(pub_key)`
function:
`...`
`return key, Capsule(point_eph_e=pub_r, point_eph_v=pub_u, bn_sig=s)`

We'll have:

caller:
"""
sym_key, capsule_elements = pre.encapsulate(pub_key)
capsule = Capsule(*capsule_elements)
"""
function:
"""
...
return key, dict(point_eph_e=pub_r, point_eph_v=pub_u, bn_sig=s)
"""

tuxxy:
I'm gonna go ahead and merge this. It seems like we have reached agreement in chat and here.

PR #24, opened 2018-01-18
[NO-MERGE] Unified Capsule instead of ReconstructedCapsule

jMyles:
### What this does:
1. Removes ReconstructedCapsule
2. Provides facilities for instantiating a Capsule from reconstructed components
3. Establishes public interfaces for Bob to open the capsule and get the contents inside.

### DIFF
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -424,7 +424,7 @@ class PRE(object):
     def decapsulate_original(self, priv_key, capsule, key_length=32):
         """Derive the same symmetric key"""
 
-        shared_key = (capsule.point_eph_e + capsule.point_eph_v) * priv_key
+        shared_key = capsule.original_ephemeral_points() * priv_key
         key = kdf(shared_key, key_length)
 
         # Check correctness of original ciphertext (check nº 2) at the end 
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -13,6 +13,7 @@ class UmbralParameters(object):
         self.h = Point.gen_rand(self.curve)
         self.u = Point.gen_rand(self.curve)
 
+
 class KFrag(object):
     def __init__(self, id_, key, x, u1, z1, z2):
         self.bn_id = id_
@@ -54,12 +55,12 @@ class KFrag(object):
         u1 = self.point_commitment
         z1 = self.bn_sig1
         z2 = self.bn_sig2
-        x  = self.point_eph_ni
+        x = self.point_eph_ni
 
         g_y = (params.g * z2) + (pub_a * z1)
 
         return z1 == hash_to_bn([g_y, self.bn_id, pub_a, pub_b, u1, x], params)
-    
+
     def is_consistent(self, vKeys, params: UmbralParameters):
         if vKeys is None or len(vKeys) == 0:
             raise ValueError('vKeys must not be empty')
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -140,8 +140,8 @@ class Capsule(object):
         self.cfrags = {}
         self._contents = None
 
-    @staticmethod
-    def from_bytes(data: bytes, curve: ec.EllipticCurve):
+    @classmethod
+    def from_original_bytes(cls, data: bytes, curve: ec.EllipticCurve):
         """
         Instantiates a Capsule object from the serialized data.
         """
@@ -149,7 +149,25 @@ class Capsule(object):
         eph_v = Point.from_bytes(data[33:66], curve)
         sig = BigNum.from_bytes(data[66:98], curve)
 
-        return Capsule(eph_e, eph_v, sig)
+        return cls(eph_e, eph_v, sig)
+
+    @classmethod
+    def from_reconstructed_bytes(cls, data: bytes, curve: ec.EllipticCurve):
+        """
+        Instantiate a Capsule from serialized data after reconstruction has occurred.
+
+        The most obvious use case is Bob affixing at least m cFrags and then serializing the Capsule.
+        """
+        # TODO: Seems like a job for BytestringSplitter ?
+        e_prime = Point.from_bytes(data[0:33], curve)
+        v_prime = Point.from_bytes(data[33:66], curve)
+        eph_ni = Point.from_bytes(data[66:99], curve)
+
+        return cls(e_prime=e_prime, v_prime=v_prime, noninteractive_point=eph_ni)
+
+    @property
+    def contents(self):
+        return self._contents
 
     def to_bytes(self):
         """
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -173,17 +173,17 @@ class Capsule(object):
         """
         Serialize the Capsule into a bytestring.
         """
-        eph_e = self.point_eph_e.to_bytes()
-        eph_v = self.point_eph_v.to_bytes()
-        sig = self.bn_sig.to_bytes()
+        eph_e = self._point_eph_e.to_bytes()
+        eph_v = self._point_eph_v.to_bytes()
+        sig = self._bn_sig.to_bytes()
 
         return eph_e + eph_v + sig
 
     def verify(self, params: UmbralParameters):
 
-        e = self.point_eph_e
-        v = self.point_eph_v
-        s = self.bn_sig
+        e = self._point_eph_e
+        v = self._point_eph_v
+        s = self._bn_sig
         h = hash_to_bn([e, v], params)
 
         return params.g * s == v + (e * h)
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -1,9 +1,9 @@
-from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import ec
 
 from umbral.bignum import BigNum
 from umbral.point import Point
 from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf
+
 
 class UmbralParameters(object):
     def __init__(self):
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -209,6 +209,9 @@ class Capsule(object):
             newly_opened = False
         return self.contents, newly_opened
 
+    def original_components(self):
+        return self._point_eph_e, self._point_eph_v, self._bn_sig
+
     def _reconstruct(self, pre):
         id_cfrag_pairs = list(self.cfrags.items())
         id_0, cfrag_0 = id_cfrag_pairs[0]
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -191,21 +191,15 @@ class Capsule(object):
     def attach_cfrag(self, cfrag: CapsuleFrag):
         self.cfrags[cfrag.bn_kfrag_id] = cfrag
 
-    def original_ephemeral_points(self):
-        return self._point_eph_e + self._point_eph_v
-
-    def reconstructed_ephemeral_points(self):
-        return self._point_eph_e_prime + self._point_eph_v_prime
-
     def open(self, pub_bob, priv_bob, pub_alice, force_reopen=False, pre=None):
-        # TODO: Raise an error here if Bob has gathered enough cFrags.
+        # TODO: Raise an error here if Bob hasn't gathered enough cFrags.
         if self.contents and not force_reopen:
             newly_opened = True
         else:
             self._reconstruct(pre=pre)
             if not pre:
                 pre = PRE(UmbralParameters())
-            self._contents = pre._decapsulate_reencrypted(pub_bob, priv_bob, pub_alice, self)
+            self._contents = pre.decapsulate_reencrypted(pub_bob, priv_bob, pub_alice, self)
             newly_opened = False
         return self.contents, newly_opened
 
@@ -427,7 +421,7 @@ class PRE(object):
     def decapsulate_original(self, priv_key, capsule, key_length=32):
         """Derive the same symmetric key"""
 
-        shared_key = capsule.original_ephemeral_points() * priv_key
+        shared_key = (capsule._point_eph_e + capsule._point_eph_v) * priv_key
         key = kdf(shared_key, key_length)
 
         # Check correctness of original ciphertext (check nº 2) at the end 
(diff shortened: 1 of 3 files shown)


umbral/umbral.py
```diff
@@ -388,36 +424,35 @@ def encapsulate(self, pub_key, key_length=32):
...
-        v = original_capsule.point_eph_v
-        s = original_capsule.bn_sig
-        h = hash_to_bn([e, v], self.params)
-        inv_d = ~d
-        assert orig_pub_key * (s * inv_d) == v_prime + (e_prime * h), "Generic Umbral Error"
+        xcomp = capsule._point_noninteractive
 
-        return key
+        d = hash_to_bn([xcomp, pub_key, xcomp * priv_key], self.params)
+        shared_key = capsule.reconstructed_ephemeral_points() * d
```
tuxxy: Don't do this, this is not clear at all with what's happening. It doesn't look like DH anymore.

umbral/umbral.py
```diff
@@ -388,36 +424,35 @@ def encapsulate(self, pub_key, key_length=32):
...
+
+        if capsule._point_eph_e:
+            # TODO: So, here, we have Alice's data too.  What's actually going on here?
+            e = capsule._point_eph_e
+            v = capsule._point_eph_v
+            s = capsule._bn_sig
+            h = hash_to_bn([e, v], self.params)
+            inv_d = ~d
+            assert orig_pub_key * (s * inv_d) == capsule._point_eph_v_prime + (
+                capsule._point_eph_e_prime * h), "Generic Umbral Error"
```
tuxxy: This is really hard to read, can you leave in the `e_prime` and `v_prime` vars?

umbral/umbral.py
```diff
@@ -388,36 +424,35 @@ def encapsulate(self, pub_key, key_length=32):
 
         return key, Capsule(point_eph_e=pub_r, point_eph_v=pub_u, bn_sig=s)
 
-
     def decapsulate_original(self, priv_key, capsule, key_length=32):
         """Derive the same symmetric key"""
 
-        shared_key = (capsule.point_eph_e + capsule.point_eph_v) * priv_key
+        shared_key = capsule.original_ephemeral_points() * priv_key
```
tuxxy: Doesn't look like it's doing DH anymore. :/

umbral/umbral.py
```diff
@@ -115,98 +116,135 @@ def __bytes__(self):
...
-        e_prime = Point.from_bytes(data[0:33], curve)
-        v_prime = Point.from_bytes(data[33:66], curve)
-        eph_ni = Point.from_bytes(data[66:99], curve)
+        eph_e = self._point_eph_e_prime
+        eph_v = self._point_eph_v_prime
+        point_noninter = self._point_noninteractive
 
-        return ReconstructedCapsule(e_prime, v_prime, eph_ni)
-
-    def to_bytes(self):
```
tuxxy: Why are you getting rid of `to_bytes`? Keep this for uniformity, it's how cryptographic APIs handle their bytestring representations.

umbral/umbral.py
```diff
@@ -115,98 +116,135 @@ def __bytes__(self):
...
-        self.point_eph_e_prime = e_prime
-        self.point_eph_v_prime = v_prime
-        self.point_eph_ni = x
+        self._point_eph_e_prime = e
+        self._point_eph_v_prime = v
+        self._point_noninteractive = cfrag_0.point_eph_ni
 
-    @staticmethod
-    def from_bytes(data: bytes, curve: ec.EllipticCurve):
+    def _reconstructed_bytes(self):
```
tuxxy: Please keep it as `from_bytes`,  we're going to also want this "public".

umbral/umbral.py
```diff
@@ -115,98 +116,135 @@ def __bytes__(self):
...
 
     def attach_cfrag(self, cfrag: CapsuleFrag):
         self.cfrags[cfrag.bn_kfrag_id] = cfrag
 
-    def reconstruct(self):
+    def original_ephemeral_points(self):
+        return self._point_eph_e + self._point_eph_v
+
+    def reconstructed_ephemeral_points(self):
+        return self._point_eph_e_prime + self._point_eph_v_prime
```
tuxxy: Please rather call these vars individually for more readability. This also does point addition, did you notice that?

umbral/umbral.py
```diff
@@ -115,98 +116,135 @@ def __bytes__(self):
...
         """
         eph_e = Point.from_bytes(data[0:33], curve)
         eph_v = Point.from_bytes(data[33:66], curve)
         sig = BigNum.from_bytes(data[66:98], curve)
 
-        return Capsule(eph_e, eph_v, sig)
+        return cls(eph_e, eph_v, sig)
+
+    @classmethod
+    def from_reconstructed_bytes(cls, data: bytes, curve: ec.EllipticCurve):
```
tuxxy: Please rename to `from_bytes` for uniformity.

umbral/umbral.py
```diff
@@ -115,98 +116,135 @@ def __bytes__(self):
...
+        return self._point_eph_e + self._point_eph_v
+
+    def reconstructed_ephemeral_points(self):
+        return self._point_eph_e_prime + self._point_eph_v_prime
+
+    def open(self, pub_bob, priv_bob, pub_alice, force_reopen=False, pre=None):
+        # TODO: Raise an error here if Bob has gathered enough cFrags.
+        if self.contents and not force_reopen:
+            newly_opened = True
+        else:
```
tuxxy: This `newly_opened` and `force_reopen` doesn't really do much for me. It just adds some weird logic in an otherwise plain reference implementation. I don't understand what problem it's solving.

umbral/umbral.py
jMyles: Well, which one is `from_bytes`?  This or `from_original_bytes`?

umbral/umbral.py
jMyles: ...why?

I don't think that three fairly obscure var names are more (or even as) readable as a single method name.

jMyles: ...why?

I don't think that three fairly obscure var names are more (or even as) readable as a single method name.

jMyles: ...why?

I don't think that three fairly obscure var names are more (or even as) readable as a single method name.

umbral/umbral.py
jMyles: Yeah I mostly agree.  I just didn't want to leave the user without this functionality without at least sketching and discussing it first.

umbral/umbral.py
tuxxy: I would remove both and leave in `from_bytes`, and offer a flag for `is_reconstructed=False`. When `True`, reconstruct the capsule.

umbral/umbral.py
jMyles: I don't understand what you mean here.  This isn't a constructor method like `from_bytes`.   Maybe you misread?

umbral/umbral.py
tuxxy: Because you're hiding point addition -- this does a few things that we don't want.
    - Since this is a reference implementation, anywhere in the code should be really explicit when we're performing a curve operation. 
    - This is a _dangerous_ operation and shouldn't be exposed as a simple method that the user can confuse for something else. This method is not clear that it performs addition, even with the operator overloading.

umbral/umbral.py
tuxxy: I think this should probably stay out of this PR until David and I discuss how we're going to handle the symmetric operations when something like this might be possible.

umbral/umbral.py
jMyles: Ahh yes that sounds good.

umbral/umbral.py
tuxxy: Ah, you're right. Probably will also want to do how you're doing `from_bytes` above with the boolean flag, except do it as `to_bytes` here.

jMyles: Sounds good.

umbral/umbral.py
tuxxy: Another issue here are the names, I just realized:

`original_ephemeral_points` -> It's not returning plural points, after the addition it returns one and they're not the original ones either. Same for `reconstructed_ephemeral_points`.

jMyles: OK, that's a good point.

I still think though that returning them from a single uniform method is much safer.

For the smaller userbase that is willing to assume the liability, the variables are available as private attributes.  But the better approach is to provide a method that does the addition, the correct way, with the correct two values, and to make that method the public default.

So: how about if these two methods instead return a tuple of the points?

This also has the benefit of forcing the candor in the implementation that you're after, ie:

`sum(capsule.original_ephemeral_points) * priv_key`

tuxxy: No, that's just clear to me. Can we just leave it as it was? It makes it so much more clear.

jMyles: I think I'm interested in other perspectives here.  @cygnusv, @michwill, @KPrasch?

What's the advantage of *forcing* the user to try to do it properly instead of giving them the method that does (but also providing the primitives)?

Consider:

`shared_key = (capsule.point_eph_e + capsule.point_eph_v) * priv_key`

Have I used the correct two points?
- Did I use the original if that's what I wanted?
- Did I use the reconstructed if that's what I wanted?
- Did I use both from the same set?  I didn't intermix, did I?
- Am I sure I didn't let the `sig` or `noninteractive_point` slip in there?  Definiely `e` and `v`, right?

...especially since there's no great way to `repr` Points from each other in such as way that makes clear their role.

Now, did I add them in the correct order?

Now consider:

`shared_key = sum(capsule.original_ephemeral_points) * priv_key`

There's no way to mess that up.  I know that I have the original points, I know that I have the correct two values, that they're in the right order, that they're added properly.  Nothing is left to chance.

What is bad about this?  Are you suggesting that other libraries refuse to do this on principle?  If so, I want to see the conversation that led to such a design decision - do you have a link?

jMyles: OK, we decided not to do this.  From Slack:

	
jmyles 5:51 PM
Getting back `original_ephemeral_points` for a minute:
5:51
Are you saying that this is never a public notion at all?
tux 5:51 PM
I don’t expect anyone to modify the underlying cryptography of this library. If they do, they probably shouldn’t.
jmyles 5:51 PM
Well, that's not what I asked.  :slightly_smiling_face:
dnunez 5:52 PM
they are in the test because we need to check the serialization
tux 5:52 PM
Right
dnunez 5:52 PM
but a user would never need to inspect those elements
jmyles 5:52 PM
I see
5:52
ok
dnunez 5:52 PM
for a user a capsule is atomic
jmyles 5:52 PM
So they are used by their respective decapsulate functions and never publicly?
5:52
Ahh
5:52
OK
dnunez 5:52 PM
yes
jmyles 5:52 PM
OK yeah, I withdraw then - I misunderstood.
dnunez 5:53 PM
we only use it in umbral, and in the tests of course
5:53
but there shouldn't be used outside that scope
jmyles 5:53 PM
I see.
5:53
Yeah, that completely makes sense now, reading again
dnunez 5:53 PM
in any case, I don't see why is this related with that method that returns the additon of e and v
jmyles 5:54 PM
Well, my drive was to help the user avoid doing that improperly.
5:54
(ie, with the original 'e' and 'v' when they wanted the reconstructed, etc)
5:55
See, this is part of the problem with writing the first time with too few `_private` things :slightly_smiling_face:
dnunez 5:55 PM
ok, that possibly was a problem, yes
tux 5:56 PM
I thought we all knew that all those fragment pieces were parts of the ciphertext?
5:56
intrinsically private sorta :sweat_smile:
dnunez 5:57 PM
it was clear to me, at least, but of course I designed it
5:57
and I have to admit that I'm new to python
tux 5:57 PM
Meh, don’t fret about Python privacy, it’s not real privacy.
dnunez 5:57 PM
the `_private` notation seems weird coming from Java
jmyles 5:57 PM
Yeah, I mean reading again, with the benefit of you're commentary, it makes sense
tux 5:58 PM
It just doesn’t import certain variables prefixed with `_`
dnunez 5:58 PM
since it's not real
jmyles 5:58 PM
But do you see what I was seeing?
dnunez 5:58 PM
not really
5:58
but I understand that there could have been misunderstandings
jmyles 5:58 PM
I saw `shared_key = (capsule.point_eph_e + capsule.point_eph_v) * priv_key` in a public method and thought, "well, if the user wanted to..."
michwill 5:58 PM
Well, you can have two underscores for real :slightly_smiling_face:
tux 5:59 PM
throws papers
jmyles 5:59 PM
hahahaha
5:59
But seriously: the decapsulate methods are good candidates for `__`
5:59
for this reason
dnunez 5:59 PM
I don't see what's the potential problem with the user anyway
tux 6:00 PM
^ Yeah, this is really my main question
6:00
I don’t expect users to go muddling around in the code here.
dnunez 6:00 PM
how could the user screw things there?
tux 6:00 PM
It’s not like they’re gonna go, “Oh, well I need to add to the ciphertext!”
jmyles 6:00 PM
Well, what if they wrote a Bob who decapsulates the capsule in his super special way, but used the original `e` and the reconstrcuted `v`?
tux 6:00 PM
_they should not do that_
dnunez 6:00 PM
basically, it would fail
jmyles 6:01 PM
@tux: if you feel so strongly, then let's have code saying so :slightly_smiling_face:
6:01
@tux: I agree *they should not do that* - that's why I wrote a method that does it the right way for them
6:01
(not yet realizing that they don't need to do it anyway)
dnunez 6:01 PM
but jmyles
6:01
your argument applies to every line of the code
jmyles 6:01 PM
No, only the public interfaces
dnunez 6:02 PM
I dont understand that
jmyles 6:02 PM
If `Capsule.open` calls `__decapsulate_reconstructed`, then it's absolutely obvious that the user isn't allowed to much with the latter without taking complete responsibility for their idiocy
6:02
:slightly_smiling_face:
dnunez 6:02 PM
ok...
jmyles 6:02 PM
I was trying to remove `ReconstructedCapsule` in a way that still forced the user to only be able to do the right thing.
6:03
And, at the time, I thought that I was seeing the user calling `decapsulate` directly, not using the public interfaces on `Capsule` (because yes, I spaced for a second and wasn't thinking of a capsule as an atomic structure)

umbral/umbral.py
```diff
@@ -115,98 +116,135 @@ def __bytes__(self):
...
+            self._reconstruct(pre=pre)
+            if not pre:
+                pre = PRE(UmbralParameters())
```
cygnusv: UmbralParameters() should not be called here. There is only ONE set of parameters for the whole system, and UmbralParameters creates these parameters, where some of them are randomized. Therefore if we create new parameters, things will start not to work. Perhaps I should have used some kind of singleton pattern for this parameters, of something similar.

jMyles: @cygnusv absolutely - I was going to make an issue for this next.

Let's turn the PRE class into a module, and instantiate the one and only `UmbralParamteres` there.

Consider this a workaround until we do that.

tuxxy:
I'm locking the conversation here. Further discussion should happen in the more recent PR #38.

This will be closed when #38 gets merged.

PR #25, opened 2018-01-18
[WIP] Better errors.

jMyles:


tuxxy:
Status on this? Are we going to drop this PR or what's happening?

jMyles:
Probably drop, but give me a few hours to mull it over.

tuxxy:
Sure, take your time. It can always be reopened.

jMyles:
OK yeah, this is an obvious drop, but I'm not ready to call it wontfix because I think that it is still a good idea to have a better error than "Generic Umbral Error" in the event that Bob doesn't have enough CFrags.

PR #26, opened 2018-01-19
Some code organization - began moving PRE to its own module.

jMyles:
PRE is still a class, but is now in its own module.


jMyles:
@cygnusv: This may or may not be a good starting point for you.  I have some other ideas about how to work out imports and where to put things, but right now I need sleep.

cygnusv:
Something I'm not comfortable with is the dismemberment of Umbral code in several files. In a previous comment with @tuxxy I told him that I personally preferred to have all (or at least most of) the code of Umbral in a single file, since it is a reference implementation and it's not that long.

Can this be compatible with the transformation of PRE to a module?

tuxxy:
When I get some time, I'll put my InfoSec code auditor hat on and compare the two and see if anything sticks out to me.

tuxxy:
However, I do share your concerns as well. I don't like hiding security critical code within language or implementation details, nor do I like complex or lots of branching logic.

jMyles:
@cygnusv:  Yeah, I think that we can keep things in a single module, but have the PRE and params be module-scoped if that's what you prefer.

When you say that you "preferred to have all (or at least most of) the code of Umbral in a single file," what exactly constitues the "code of Umbral" in your mind?  You already moved Point and BigNum out.  I imagined that PRE might constitute the "core reference", with only bytes transiting in and out for auditability, and then the other modules having the classes that get instantiated with the elements returns by the PRE functions.

tuxxy:
Status on this? I think it was decided that it shouldn't be its own module. @jMyles Are we good to close this?

jMyles:
@tuxxy I had planned to close it after disposition on #38, but there's actually no reason to do that now, since this will clearly not be merged.

However, I am still of the mind that breaking up umbral.py is a very, very good decision.

tuxxy:
It can always be re-opened in the future. <3 Github

PR #27, opened 2018-01-19
This is just #23 with a quick merge.

jMyles:


tuxxy:
Oh, I didn't realize this was a thing.. Does this need to remain open for any reason?

PR #29, opened 2018-01-23
Implement Simple API to PRE and add UmbralDEM class

tuxxy:
### What this does:
1. Implements a simple API with the following methods:
   - `encrypt` -- Encrypts data using KEM/DEM.
   - `decrypt` -- Decrypts data using KEM/DEM.
   - `decrypt_reencrypted` -- Decrypts threshold encrypted data using KEM/DEM.
   - `reencrypt` -- Already in the `PRE` class.
   - `split_rekey` -- Already in the `PRE` class. NOTE: Returns `vKeys`, disregard this for the time being via `_` operator.
   - `reconstruct` -- Already in the `Capsule` class. It is required to use the above methods; therefore is exposed in a similar fashion already.
2. Modifies input of the `split_rekey` method to allow for `UmbralPrivateKey` and `UmbralPublicKey` use.
3. Adds `gen_key` method to `UmbralPrivateKey`.
   - Adds method `get_pub_key` to `UmbralPrivateKey` to calculate public key for the private key.
4. Privatizes the following methods in `PRE`:
    - `_encapsulate`
    - `_decapsulate_reencrypted`
5. Adds `UmbralDEM`:
    - Handles symmetric encryption using NaCl's Salsa20-Poly1305.
         - Maybe switch to a cipher that allows authenticated data? This needs to be scoped/researched.
    - Left to the user to read data from filesystem.
6. Fixes a circular import with `UmbralParameters`.
7. Adds tests for all these things.

---

@cygnusv and I had a call around 02:00MST/01:00PST and discussed the needs for the `UmbralDEM` class as well as the simple API implemented here.

The simple API is implemented in the core Umbral class `PRE`. Subpaths need to be handled on the KMS side. This does not implement or use any challenge methods.

##### NOTE: This merges to the `kms-depend` branch. After merging this branch, continue to merge to the `master` branch.

Good night.

### DIFF
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -45,7 +45,7 @@ def test_simple_api(N, threshold):
     dec_data = pre.decrypt(priv_key_alice, capsule, enc_data)
     assert dec_data == plain_data
 
-    rekeys, _ = pre.split_rekey(priv_key_alice, pub_key_bob, threshold, N)
+    rekeys, _unused_vkeys = pre.split_rekey(priv_key_alice, pub_key_bob, threshold, N)
     for rekey in rekeys:
         cFrag = pre.reencrypt(rekey, capsule)
         capsule.attach_cfrag(cFrag)
@@ -91,7 +91,7 @@ def test_kfrag_serialization():
     priv_key = pre.gen_priv()
     pub_key = pre.priv2pub(priv_key)
 
-    kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
+    kfrags, _unused_vkeys = pre.split_rekey(priv_key, pub_key, 1, 2)
     kfrag_bytes = kfrags[0].to_bytes()
 
     # A KFrag can be represented as the 194 total bytes of two Points (33 each) and four BigNums (32 each).
@@ -113,8 +113,8 @@ def test_cfrag_serialization():
     priv_key = pre.gen_priv()
     pub_key = pre.priv2pub(priv_key)
 
-    _, capsule = pre._encapsulate(pub_key)
-    kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
+    _unused_key, capsule = pre._encapsulate(pub_key)
+    kfrags, _unused_vkeys = pre.split_rekey(priv_key, pub_key, 1, 2)
 
     cfrag = pre.reencrypt(kfrags[0], capsule)
     cfrag_bytes = cfrag.to_bytes()
@@ -156,8 +156,8 @@ def test_reconstructed_capsule_serialization():
     priv_key = pre.gen_priv()
     pub_key = pre.priv2pub(priv_key)
 
-    _, capsule = pre._encapsulate(pub_key)
-    kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
+    _unused_key, capsule = pre._encapsulate(pub_key)
+    kfrags, _unused_vkeys = pre.split_rekey(priv_key, pub_key, 1, 2)
 
     cfrag = pre.reencrypt(kfrags[0], capsule)
 
@@ -183,8 +183,8 @@ def test_challenge_response_serialization():
     priv_key = pre.gen_priv()
     pub_key = pre.priv2pub(priv_key)
 
-    _, capsule = pre._encapsulate(pub_key)
-    kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
+    _unused_key, capsule = pre._encapsulate(pub_key)
+    kfrags, _unused_vkeys = pre.split_rekey(priv_key, pub_key, 1, 2)
 
     cfrag = pre.reencrypt(kfrags[0], capsule)
 
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -33,25 +33,25 @@ def test_simple_api(N, threshold):
     params = umbral.UmbralParameters()
     pre = umbral.PRE(params)
 
-    priv_key = keys.UmbralPrivateKey.gen_key(params)
-    pub_key = priv_key.get_pub_key(params)
+    priv_key_alice = keys.UmbralPrivateKey.gen_key(params)
+    pub_key_alice = priv_key_alice.get_pub_key(params)
 
     priv_key_bob = keys.UmbralPrivateKey.gen_key(params)
     pub_key_bob = priv_key_bob.get_pub_key(params)
 
     plain_data = b'attack at dawn'
-    enc_data, capsule = pre.encrypt(pub_key, plain_data)
+    enc_data, capsule = pre.encrypt(pub_key_alice, plain_data)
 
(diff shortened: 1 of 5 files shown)


tests/test_umbral.py
```diff
@@ -18,15 +19,43 @@ def test_decapsulation_by_alice():
...
 
 
+@pytest.mark.parametrize("N,threshold", parameters)
+def test_simple_api(N, threshold):
+    params = umbral.UmbralParameters()
+    pre = umbral.PRE(params)
+
+    priv_key = keys.UmbralPrivateKey.gen_key(params)
+    pub_key = priv_key.get_pub_key(params)
+
```
cygnusv: Perhaps change `priv_key` to `priv_key_alice`. The same for `pub_key`.

tests/test_umbral.py
```diff
@@ -18,15 +19,43 @@ def test_decapsulation_by_alice():
...
+    rekeys, _ = pre.split_rekey(priv_key, pub_key_bob, threshold, N)
+    for rekey in rekeys:
+        cFrag = pre.reencrypt(rekey, capsule)
+        capsule.attach_cfrag(cFrag)
+
+    reenc_dec_data = pre.decrypt_reencrypted(
+        priv_key_bob, pub_key_bob, pub_key, capsule, enc_data
+    )
+    assert reenc_dec_data == plain_data
+
```
cygnusv: Great, this is completely in line with Michael's API.

umbral/umbral.py
```diff
@@ -421,3 +436,55 @@ def decapsulate_reencrypted(self, pub_key: Point, priv_key: BigNum, orig_pub_key
...
+                            recp_pub_key: UmbralPublicKey,
+                            sender_pub_key: UmbralPublicKey,
+                            capsule: Capsule, enc_data: bytes):
+        """
+        Decrypts the data provided by reconstructing and decapsulating the
+        capsule.
+
+        Returns the plaintext of the data.
+        """
+        # TODO: Do we take only the recipient's private key and form pubkey?
```
cygnusv: Yes, we can do that, since the public key can be deterministically derived from the private key.

tests/test_umbral.py
```diff
@@ -18,15 +19,43 @@ def test_decapsulation_by_alice():
...
 
-    sym_key_2 = pre.decapsulate_original(priv_key, capsule)
+    sym_key_2 = pre._decapsulate_original(priv_key, capsule)
     assert sym_key_2 == sym_key
 
 
+@pytest.mark.parametrize("N,threshold", parameters)
+def test_simple_api(N, threshold):
+    params = umbral.UmbralParameters()
+    pre = umbral.PRE(params)
```
jMyles: Just getting going on a more thorough review.  Overall, this is a *huge* improvement from 24 hours ago!

I'll make Issues for things that are, at least to me, outstanding.  I started on this line, with #31.

tests/test_umbral.py
```diff
@@ -18,15 +19,43 @@ def test_decapsulation_by_alice():
...
+    plain_data = b'attack at dawn'
+    enc_data, capsule = pre.encrypt(pub_key_alice, plain_data)
+
+    dec_data = pre.decrypt(priv_key_alice, capsule, enc_data)
+    assert dec_data == plain_data
+
+    rekeys, _ = pre.split_rekey(priv_key_alice, pub_key_bob, threshold, N)
+    for rekey in rekeys:
+        cFrag = pre.reencrypt(rekey, capsule)
+        capsule.attach_cfrag(cFrag)
```
michwill: Ah interesting. So we attach cFrag to exactly the original capsule, not a brand new one?

tests/test_umbral.py
```diff
@@ -18,15 +19,43 @@ def test_decapsulation_by_alice():
...
+    pre = umbral.PRE(params)
+
+    priv_key_alice = keys.UmbralPrivateKey.gen_key(params)
+    pub_key_alice = priv_key_alice.get_pub_key(params)
+
+    priv_key_bob = keys.UmbralPrivateKey.gen_key(params)
+    pub_key_bob = priv_key_bob.get_pub_key(params)
+
+    plain_data = b'attack at dawn'
+    enc_data, capsule = pre.encrypt(pub_key_alice, plain_data)
```
jMyles: So, we really don't want to allow the user to optionally use their own symmetric tooling here?  I think this is a *big* decision decision to be made without discussion.

Although, maybe it makes sense to do it this way for now and add the option later if we want?

tests/test_umbral.py
```diff
@@ -84,7 +113,7 @@ def test_cfrag_serialization():
     priv_key = pre.gen_priv()
     pub_key = pre.priv2pub(priv_key)
 
-    _, capsule = pre.encapsulate(pub_key)
+    _, capsule = pre._encapsulate(pub_key)
```
jMyles: I don't really like `_` as a name in tests - do you think it makes sense to just call this `_key_bytes`?

tests/test_umbral.py
```diff
@@ -84,7 +113,7 @@ def test_cfrag_serialization():
...
     kfrags, _ = pre.split_rekey(priv_key, pub_key, 1, 2)
```
jMyles: Same here - maybe call this `_verification_keys`.

tests/test_umbral.py
michwill: I think, this can eventually be handled by having a default dem as a class attribute (well, at least on the basic level).
E.g.: `dem = UmbralDEM(...)` ->
```
class PRE(object):
    ...
    DEM = UmbralDEM
```
 ... `dem = PRE.DEM(...)`

tests/test_umbral.py
tuxxy: You attach it, then call `capsule.reconstruct` which returns a new capsule.

tests/test_umbral.py
tuxxy: It would probably be for the best, currently, if the user can't easily extend their own DEM. Umbral is defined to use Salsa20-Poly1305, but we'll be switching to ChaCha20-Poly1305 to allow for us to add in authenticated data for more security in the overall system.

For that reason, users really shouldn't try to roll their own crypto here.

michwill: Well yeah, I think this adds +1 to having `cls.DEM`, but will wait on @jMyles opinion

jMyles: I prefer for (in fact, I think that it only makes sense for) PRE to be a module, not a class.  So maybe instead we have:

`pre.set_dem_interface(MyCustomDEMClass)`

tuxxy: We really need the KMS to have one standard DEM method. There isn't any good reason to have to make anything more complex when we have figured the one great way to do it.

jMyles: @tuxxy: Oh yeah, I think we all agree there.  I think this concern is more about making sure we have flexibility to build things other than the KMS.  

The way you've written this, I think we can probably just add an additional interface later without too much trouble.  So I think we can actually just move with this as-is.

PR #33, opened 2018-01-24
Implement cryptography.io's AEAD for ChaCha20-Poly1305 in the UmbralDEM

tuxxy:
### What this does:
1. Replaces `pynacl` as a dependency, and uses cryptography.io's ChaCha20-Poly1305 AEAD construct.
    - Allows us to use authenticated data in the encryption/decryption param `authenticated_data`.
2. Adds tests for `UmbralDEM` as `test_dem.py`.
3. Adds constants in `dem.py` for keysize and nonce:
    - `DEM_KEYSIZE = 32`
    - `DEM_NONCE_SIZE = 12`

### DIFF
--- a/tests/test_dem.py
+++ b/tests/test_dem.py
@@ -1,0 +1,60 @@
+import pytest
+import os
+
+from umbral.dem import UmbralDEM
+from cryptography.exceptions import InvalidTag
+
+
+def test_encrypt_decrypt():
+    key = os.urandom(32)
+
+    dem = UmbralDEM(key)
+
+    plaintext = b'attack at dawn'
+
+    ciphertext0 = dem.encrypt(plaintext)
+    ciphertext1 = dem.encrypt(plaintext)
+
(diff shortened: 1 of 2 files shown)


tests/test_dem.py
```diff
@@ -0,0 +1,60 @@
...
+    ciphertext1 = dem.encrypt(plaintext)
+
+    assert ciphertext0 != plaintext
+    assert ciphertext1 != plaintext
+
+    # Ciphertext should be different even with same plaintext.
+    assert ciphertext0 != ciphertext1
+
+    # Nonce should be different
+    assert ciphertext0[:12] != ciphertext1[:12]
```
cygnusv: Perhaps for clarity you can use DEM_KEYSIZE and DEM_NONCE_SIZE to replace 32 and 12, respectively.

cygnusv: And a comment explaining that the ciphertext format is `nonce + enc_data + tag`

Issue #35, opened 2018-01-25
Deterministically generate parameters `h` and `u`


tuxxy:
We need to deterministically generate the Umbral parameters `h` and `u`.

tuxxy:
Resolved in #41.

PR #41, opened 2018-01-30
Deterministic generation of parameters h and u

cygnusv:
# What this does:

1. It defines a function called  `unsafe_hash_to_point` that hashes arbitrary data to EC points, using the try-and-increment method. Note that this method is not secure when the data is secret, since it leaks the timing.
2. It uses `unsafe_hash_to_point` to deterministically generate Umbral parameters `u` and `h`, so that both are bound to the initial generator `g`, but their discrete logarithm wrt to `g` is unknown (assuming `unsafe_hash_to_point` behaves like a random oracle). This solves issue #35 .
3. Add test for compliance of SECP256k1 generator (i.e., our current `g`) with respect to the SEC standard

### DIFF
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -16,11 +16,10 @@ class UmbralParameters(object):
         
         g_bytes = self.g.to_bytes(is_compressed=True)
 
-        constant_u = b'NuCypherKMS/UmbralParameters/u'
-        constant_h = b'NuCypherKMS/UmbralParameters/h'
+        domain_seed = b'NuCypherKMS/UmbralParameters/'
 
-        self.h = unsafe_hash_to_point(self.curve, g_bytes, constant_h)
-        self.u = unsafe_hash_to_point(self.curve, g_bytes, constant_u)
+        self.h = unsafe_hash_to_point(self.curve, g_bytes, domain_seed + b'h')
+        self.u = unsafe_hash_to_point(self.curve, g_bytes, domain_seed + b'u')
 
 class KFrag(object):
     def __init__(self, id_, key, x, u1, z1, z2):
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -5,7 +5,7 @@ from umbral.bignum import BigNum
 from umbral.dem import UmbralDEM
 from umbral.keys import UmbralPrivateKey, UmbralPublicKey
 from umbral.point import Point
-from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf, hash_to_point
+from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf
 
 
 class UmbralParameters(object):
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -5,7 +5,7 @@ from umbral.bignum import BigNum
 from umbral.dem import UmbralDEM
 from umbral.keys import UmbralPrivateKey, UmbralPublicKey
 from umbral.point import Point
-from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf
+from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf, unsafe_hash_to_point
 
 
 class UmbralParameters(object):
@@ -19,8 +19,8 @@ class UmbralParameters(object):
         constant_u = b'NuCypherKMS/UmbralParameters/u'
         constant_h = b'NuCypherKMS/UmbralParameters/h'
 
-        self.h = hash_to_point(self.curve, g_bytes, constant_h)
-        self.u = hash_to_point(self.curve, g_bytes, constant_u)
+        self.h = unsafe_hash_to_point(self.curve, g_bytes, constant_h)
+        self.u = unsafe_hash_to_point(self.curve, g_bytes, constant_u)
 
 class KFrag(object):
     def __init__(self, id_, key, x, u1, z1, z2):
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -6,7 +6,7 @@ from umbral.bignum import BigNum
 from umbral.point import Point
 from umbral.keys import UmbralPrivateKey, UmbralPublicKey
 from umbral.dem import UmbralDEM
-from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf
+from umbral.utils import poly_eval, lambda_coeff, hash_to_bn, kdf, hash_to_point
 
 
 class UmbralParameters(object):
@@ -14,9 +14,14 @@ class UmbralParameters(object):
         self.curve = ec.SECP256K1()
         self.g = Point.get_generator_from_curve(self.curve)
         self.order = Point.get_order_from_curve(self.curve)
-        self.h = self.g
-        self.u = self.g
+        
+        g_bytes = self.g.to_bytes(is_compressed=True)
 
+        constant_u = b'NuCypherKMS/UmbralParameters/u'
+        constant_h = b'NuCypherKMS/UmbralParameters/h'
+
+        self.h = hash_to_point(self.curve, g_bytes, constant_h)
+        self.u = hash_to_point(self.curve, g_bytes, constant_u)
 
 class KFrag(object):
     def __init__(self, id_, key, x, u1, z1, z2):
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -1,11 +1,10 @@
-from nacl.secret import SecretBox
-from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import ec
+from nacl.secret import SecretBox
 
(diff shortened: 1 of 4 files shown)


tuxxy:
@michwill I'm fairly certain an attack like that (if you mean length extension attacks) isn't quite applicable here unless this is used to hash secret values which @cygnusv points out.

The major attack application here would be when someone is using this function specifically for mapping, say, a MAC to the curve. I don't see a reason why someone would need to do that, though.

To get around all these concerns, it could be switched to SHA3, but I don't think that's needed here.

umbral/utils.py
```diff
@@ -55,6 +55,42 @@ def hash_to_bn(list, params):
  
     return res
 
+def hash_to_point(curve, data, constant=None):
```
tuxxy: Let's change this name to `unsafe_hash_to_point`. As you point out, it isn't in constant time and we should be up front about the unsafe nature of this in the name as well as the code, too.

umbral/utils.py
```diff
@@ -55,6 +55,42 @@ def hash_to_bn(list, params):
...
+        ibytes = i.to_bytes(4, byteorder='big')
+        digest = hashes.Hash(hashes.SHA256(), backend=default_backend())
+        digest.update(constant + ibytes + data)
+        hash = digest.finalize()
+
+        compressed02 = b"\x02"+hash
+        try:
+            h = Point.from_bytes(compressed02, curve)
+            return h
+        except:
```
jMyles: This is not a good approach for this situation.  Which exceptions are we actually catching here?

https://realpython.com/blog/python/the-most-diabolical-python-antipattern/

umbral/umbral.py
```diff
@@ -6,17 +6,22 @@
...
     def __init__(self):
         self.curve = ec.SECP256K1()
         self.g = Point.get_generator_from_curve(self.curve)
         self.order = Point.get_order_from_curve(self.curve)
-        self.h = self.g
-        self.u = self.g
+        
+        g_bytes = self.g.to_bytes(is_compressed=True)
+
+        constant_u = b'NuCypherKMS/UmbralParameters/u'
```
jMyles: Any interest in naming this something like, "domain"?  

`constant_u` doesn't really explain *why* this is here.

umbral/utils.py
```diff
@@ -55,6 +55,42 @@ def hash_to_bn(list, params):
...
+            pass
+
+        i+=1
+
+    # Only happens with probability 2^(-32)
+    raise ValueError('Could not find %c in %s' % (ch,str))
```
tuxxy: Let's do the Python3 specific method for string formatting:

`raise ValueError('Could not find {} in {}'.format(ch, str))`

Also, I don't see `ch` or `str` defined in this part. What's going on here?

umbral/utils.py
```diff
@@ -55,6 +55,42 @@ def hash_to_bn(list, params):
...
+    """
+    Hashes arbitrary data into a valid EC point of the specified curve,
+    using the try-and-increment method.
+    It admits an optional constant as an additional input to the hash function.
+    It uses SHA256 as the internal hash function. 
+
+    WARNING: Do not use when the input data is secret, as this implementation is not 
+    in constant time, and hence, it is not safe with respect to timing attacks.
+
+    TODO: Check how to uniformly generate ycoords. Currently, it only outputs points
```
tuxxy: Probably could do this by generating the noncompressed point via SHA512, actually.

umbral/utils.py
jMyles: :+1:

umbral/utils.py
tuxxy: Actually, yeah, I think this would work very well. @cygnusv What're your thoughts on this?

umbral/utils.py
michwill: Yeah, if there are ever any exceptions, we'd need to figure out which are those (and catch specifically them if we have to)

tuxxy: I believe it throws `cryptography.exceptions.InternalError`

umbral/umbral.py
tuxxy: It makes sense to me, but it could be renamed to `constant_u_seed` or something similar. `domain` doesn't make much sense to me.

jMyles: I don't love that either.  It's not any more "constant" than any other bytes argument to a function.

As for the "domain" idea: this is the seed (that part is good) that places this `UmbralParameters` as within the same domain as all other instances with like parameters.  So, it really is a domain denominator.

`domain_seed`?

I also don't see a good reason to have the user set both values; just have them give one and we'll append to it in order to make two.

umbral/utils.py
tuxxy: And also a potential `ValueError` from `Point.from_bytes`.

umbral/utils.py
cygnusv: Yep, I was thinking the same just now

umbral/utils.py
cygnusv: But then the ycoord is not related to the xcoord. I don't follow you

tuxxy: Oh, right. I forgot about it being a Galois field. I'll think more on this.

umbral/utils.py
cygnusv: I agree there must be a better way to do this that I don't know. My intention is 'try to deserialize this potential point, if something fails, then the point is garbage'. I really don't care what's the exception. I just know that if the potential point was valid, then there will be no exception. Feel free to suggest something along this line.

tuxxy: Would it not work to use `g` as a scalar for a hashed object?

Essentially:
`u = hash_to_bn(some_data) * g`

There might be some security implications with that... Not sure, need to look into this.

umbral/utils.py
cygnusv: The method is to for a given xcoord, compute the ycoord according to the curve equation and check if its a quadratic residue. I didn't want to do that, so I'm just using the fact that OpenSSL deserialization of compressed points checks everything implicitly, and if it fails, then some exception is raised.

umbral/utils.py
jMyles: > I really don't care what's the exception. 

The thing is: you *do* care about strange or wacky exceptions, like `RuntimeError`, `MemoryError`, or `OSError` - you don't want to accidentally catch those.

cygnusv: No, that's not a good approach. We want to give assurance that we don't know the discrete log of u wrt to g.

cygnusv: Jmyles, you're right. Didn't think about that.

michwill: @cygnusv @jMyles that's why it's so important to have both scientists and engineers here!

jMyles: How about just plain ole' humans? ;-)

cygnusv: Scientists and engineers are a good mixture :-)

umbral/umbral.py
cygnusv: So you mean we use `domain_seed + b'h'` and `domain_seed + b'u'`, right?

jMyles: Yeah, if `b'h'` and `b'u'` are the distinguishing tokens you want to use.

cygnusv: Sounds good.

umbral/utils.py
jMyles: Well, let me remind you that I am neither.

I engage in a lot of science, and a lot of engineering, but I'm not more a scientist or engineer than I am a school bus driver.

If you want to appellate y'all selves, that's fine, but I never like being excluded from the pool of people who can tackle a problem because of the perception that I'm in the wrong `class` of human.  I'm down for *any* problem - especially if it's a serious challenge - whether it's science, engineering, or bus driving.

My resume lists my profession as "Chocobo Breeder."

:-)

I will concur, though, that this PR (and also #38 / #24) have brought out of a lot of the most developed strengths in each of the four of us!

tuxxy:
@cygnusv Another question: do we have any idea what performance may look like (in worst cases) for `hash_to_point`?

cygnusv:
This try-and-increment method is probabilistic. The probability that you find a valid point is 1/2 in one try. So, in practice, you just have to try a few times on each call. Anyway, this would be used only for generating `h` and `u`. 

In fact, if performance is a concern, we could simply run this algorithm to get `h` and `u` and hard-code them, or put them in a file, or whatever. The generation procedure is there basically as an assurance that we didn't cook `h` and `u` with a backdoor or something, not necessarily because you have to compute them every time you init a PRE.

michwill:
Oh, that's a very good point. Perhaps, we should consider caching UmbralParameters in a config file, for example (maybe outside Umbral though)

cygnusv:
I just checked, and currently, using the `domain_seed = b'NuCypherKMS/UmbralParameters/'`, it takes 3 tries to compute u, and 2 for h.

PR #43, opened 2018-02-01
Have Umbral keys not require UmbralParameters to be passed

tuxxy:
## What this does:
1. Moves `UmbralParameters` to its own file `params.py`.
2. Uses default `UmbralParameters` object in the `keys` API (if none provided) for generating keypairs.
3. Adds tests for the above changes.

### DIFF
--- a/tests/test_keys.py
+++ b/tests/test_keys.py
@@ -1,47 +1,46 @@
 import pytest
 
 from umbral import umbral, keys
+from umbral.params import UmbralParameters
 
 
 def test_gen_key():
-    umbral_priv_key = keys.UmbralPrivateKey.gen_key(umbral.UmbralParameters())
+    # Pass in the parameters to test that manual param selection works
+    umbral_priv_key = keys.UmbralPrivateKey.gen_key(UmbralParameters())
     assert type(umbral_priv_key) == keys.UmbralPrivateKey
 
-    umbral_pub_key = umbral_priv_key.get_pub_key(umbral.UmbralParameters())
+    umbral_pub_key = umbral_priv_key.get_pub_key(UmbralParameters())
     assert type(umbral_pub_key) == keys.UmbralPublicKey
 
 
 def test_private_key_serialization():
-    pre = umbral.PRE(umbral.UmbralParameters())
+    pre = umbral.PRE()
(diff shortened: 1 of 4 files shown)


tests/test_keys.py
```diff
@@ -1,47 +1,46 @@
...
 
     assert priv_key == decoded_key.bn_key
 
 
 def test_public_key_serialization():
-    pre = umbral.PRE(umbral.UmbralParameters())
+    pre = umbral.PRE()
 
     priv_key = pre.gen_priv()
     pub_key = pre.priv2pub(priv_key)
```
cygnusv: There are two ways to generate public keys from private keys: using pre.priv2pub() and priv_key.gen_pub_key(). We should probably provide a single implementation, even if we maintain both functions.

cygnusv: Of course, this is not a thing to do right now. We can add it to the TODO. 
BTW, I prefer to keep priv_key.gen_pub_key() and to remove pre.priv2pub().

umbral/keys.py
```diff
@@ -87,10 +94,13 @@ def save_key(self, password: bytes=None, _scrypt_cost: int=20):
...
         return encoded_key
 
-    def get_pub_key(self, params):
+    def get_pub_key(self, params: UmbralParameters=None):
         """
         Calculates and returns the public key of the private key.
         """
+        if params is None:
+            params = UmbralParameters()
+
```
cygnusv: I think we should probably have params as an attribute of Public and Private keys. Consider this:
I create a private key using gen_key with a set of params, and then call get_pub_key with a different set of params. Of course we are not considering different sets of params, but it feels that there is too much freedom in there.

tuxxy: Yeah, you're right. I'll switch to this.

tuxxy:
@cygnusv Updates are done.

tests/test_keys.py
tuxxy: Specifically, those methods deal with `Point` and `BigNum` objects.

Yeah, it's a little weird. We could add in checks and a bool flag for the Umbral key objects there and generate accordingly.

Could also just remove them, too.

PR #47, opened 2018-02-02
Changing "_reconstruct" to "_reconstruct_shamirs_secret" and all instances of "reconstructed" to "activated".

jMyles:
The `Capsule` metaphor has metaphor has evolved a bit.  We originally conceived that the `Capsule` is shattered into fragments and had to be "recontructed."  Now, however, the story is more like this: `Bob` carries the `Capsule` around with him as he follows the `TreasureMap`, showing it to different `Ursulas` and collecting fragments.  He plugs those fragments into the `Capsule` and *it activates* - comes alive, glowing an orange-ish color so that he can provide his key and complete the decryption.

### DIFF
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -86,11 +86,11 @@ def test_two_unequal_capsules():
 
     assert one_capsule != another_capsule
 
-    reconstructed_capsule = Capsule(e_prime=Point.gen_rand(curve=pre.params.curve),
+    activated_capsule = Capsule(e_prime=Point.gen_rand(curve=pre.params.curve),
                                     v_prime=Point.gen_rand(curve=pre.params.curve),
                                     noninteractive_point=Point.gen_rand(curve=pre.params.curve))
 
-    assert reconstructed_capsule != one_capsule
+    assert activated_capsule != one_capsule
 
 
 @pytest.mark.parametrize("N,threshold", parameters)
@@ -117,7 +117,7 @@ def test_m_of_n(N, threshold):
     # assert capsule.is_openable_by_bob()  # TODO: Is it possible to check here if >= m cFrags have been attached?
     # capsule.open(pub_bob, priv_bob, pub_alice)
 
-    capsule._reconstruct()
+    capsule._activate()
     sym_key_from_capsule = pre.decapsulate_reencrypted(pub_bob, priv_bob, pub_alice, capsule)
     assert sym_key == sym_key_from_capsule
 
@@ -186,7 +186,7 @@ def test_capsule_serialization():
     # First, the public approach for the Capsule.  Simply:
     new_capsule == capsule
 
-    # Second, we show that the original components (which is all we have here since we haven't reconstructed) are the same:
+    # Second, we show that the original components (which is all we have here since we haven't activated) are the same:
     assert new_capsule.original_components() == capsule.original_components()
 
     # Third, we can directly compare the private original component attributes (though this is not a supported approach):
@@ -195,7 +195,7 @@ def test_capsule_serialization():
     assert new_capsule._bn_sig == capsule._bn_sig
 
 
-def test_reconstructed_capsule_serialization():
+def test_activated_capsule_serialization():
     pre = umbral.PRE(umbral.UmbralParameters())
 
     priv_key = pre.gen_priv()
@@ -208,10 +208,10 @@ def test_reconstructed_capsule_serialization():
 
     capsule.attach_cfrag(cfrag)
 
-    capsule._reconstruct()
+    capsule._activate()
     rec_capsule_bytes = capsule.to_bytes()
 
-    # A reconstructed Capsule is:
+    # An activated Capsule is:
     # three points, representable as 33 bytes each (the original), and
     # two points and a bignum (32 bytes) (the activated components), for 197 total.
     assert len(rec_capsule_bytes) == (33 * 3) + (33 + 33 + 32)
@@ -223,7 +223,7 @@ def test_reconstructed_capsule_serialization():
     # Again, the same three perspectives on equality. 
     new_rec_capsule == capsule
 
-    assert new_rec_capsule.reconstructed_components() == capsule.reconstructed_components()
+    assert new_rec_capsule.activated_components() == capsule.activated_components()
 
     assert new_rec_capsule._point_eph_e_prime == capsule._point_eph_e_prime
     assert new_rec_capsule._point_eph_v_prime == capsule._point_eph_v_prime
@@ -288,7 +288,7 @@ def test_challenge_response_serialization():
 #     # Let's put the re-encryption of a different Alice ciphertext
 #     cfrags[0] = pre.reencrypt(kfrags[0], other_capsule_alice)
 
-#     capsule_bob = pre.reconstruct_capsule(cfrags)
+#     capsule_bob = pre.activate_capsule(cfrags)
 
 #     try:
 #         # This line should always raise an AssertionError ("Generic Umbral Error")
@@ -333,7 +333,7 @@ def test_challenge_response_serialization():
 #     cfrags[0].point_eph_v1 = Point.gen_rand(pre.curve)
 
 
-#     capsule_bob = pre.reconstruct_capsule(cfrags)
+#     capsule_bob = pre.activate_capsule(cfrags)
 
 #     try:
 #         # This line should always raise an AssertionError ("Generic Umbral Error")
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -117,7 +117,7 @@ def test_m_of_n(N, threshold):
     # assert capsule.is_openable_by_bob()  # TODO: Is it possible to check here if >= m cFrags have been attached?
     # capsule.open(pub_bob, priv_bob, pub_alice)
 
-    capsule._activate()
+    capsule._recontruct_shamirs_secret()
     sym_key_from_capsule = pre.decapsulate_reencrypted(pub_bob, priv_bob, pub_alice, capsule)
     assert sym_key == sym_key_from_capsule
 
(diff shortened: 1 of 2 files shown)


jMyles:
The `Capsule` metaphor has metaphor has evolved a bit.  We originally conceived that the `Capsule` is shattered into fragments and had to be "recontructed."  Now, however, the story is more like this: `Bob` carries the `Capsule` around with him as he follows the `TreasureMap`, showing it to different `Ursulas` and collecting fragments.  He plugs those fragments into the `Capsule` and *it activates* - comes alive, glowing an orange-ish color so that he can provide his key and complete the decryption.

jMyles:
Some discussion: 

tux [4:22 PM]
@jmyles Not sure if I like your PR. Technically, the fragments are reconstructed.
See: https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing#Computationally_Efficient_Approach
WikipediaWikipedia
Shamir's Secret Sharing
Shamir's Secret Sharing is an algorithm in cryptography created by Adi Shamir. It is a form of secret sharing, where a secret is divided into parts, giving each participant its own unique part, where some of the parts or all of them are needed in order to reconstruct the secret.
Counting on all participants to combine the secret might be impractical, and therefore sometimes the threshold scheme is used where any
https://upload.wikimedia.org/wikipedia/commons/c/c3/Polynomial_over_a_finite_field_as_per_SSSS.png
tux [4:23 PM]
Ah, it even states it in the article:
… where some of the parts or all of them are needed in order to reconstruct the secret.
jmyles [4:25 PM]
In our case, `Bob` needs the original data of the `Capsule` in order to get the secret.  It's never shattered into pieces.  There's no moment when he needs to like reach in his pocket for glue or something.
The reconstruction metaphor that's typical of Shamir secret sharing isn't *per se* exactly the right thing for us, because our metaphor is one of encapsulation, not merely transit of a secret.
KPrasch [4:26 PM]
Yea - the language "reconstruct" almost indicates that it is a secondary (or alternate) method (edited)
especially if it is what the user is supposed to do
tux [4:27 PM]
But… we exactly use SSSS, how is it not?
jmyles [4:28 PM]
Consider: if `Bob` has < `m` fragments, then what is the `Capsule` that he's he's carting around to `Ursula` - is it broken in some way?  No, it's just as `Alice` provided it.
KPrasch [4:28 PM]
reconstruct certainly indicates that the capsule itself is in pieces
jmyles [4:28 PM]
@tux: Sure - and the fragments are an homage to that (as are our use of `m` and `n`).  But we aren't *only* doing Shamir secret sharing.  We're doing it as a means to transit a `Capsule`. (edited)
KPrasch [4:29 PM]
this is more like separate entities that interact with a capsule
tux [4:29 PM]
This just doesn’t make sense to me. “glowing an orange-ish color” what does that even mean?
tux [4:30 PM]
Activation, to me, implies that data is already somewhere and it just needs to be approved by some central authority.
It doesn’t imply changing or transformation at all
KPrasch [4:31 PM]
Fair enough
some state change needs to be named when all frags are attached to a capsule
and reconstruct seems misleading because of the points above ^
jmyles [4:34 PM]
Take some time to read the code and see if you think it reads better.  I *definitely* do.
michwill [4:37 PM]
Anyway. As a non native speaker, I am ok with reconstruct or activate. Or even combine
jmyles [4:38 PM]
Here's an angle that might be more convincing, @tux:
Consider the past tense.  A "reconstructed" `Capsule` vs. an "activated" `Capsule`.
tux [4:39 PM]
What about them?
Have to switch to mobile, give me a sec
jmyles [4:40 PM]
The former suggests that the "original" components (`_point_eph_e`, `_point_eph_v`, and `bn_sig`) are somehow transformed - perhaps broken apart and remade.
(connectivity issues - that's why things are coming in slowly)
But that's not what happens at all.
Instead, those parts endure, and *new* parts come online.
KPrasch [4:43 PM]
I have to ask myself is the capsule itself in pieces? (ie. are we building the capsule itself) (edited)
jmyles [4:43 PM]
Right - picturing the `Capsule` in pieces is likely to confuse.
dnunez [4:45 PM]
Before including cfrags in capsule it made sense to reconstruct
Now I agree that reconstruct sounds weird
But only because of the metaphor we are using
Anyway I'm not a native speaker so I don't get the nuances of both terms
And also I'm tired of metaphorical discussions, so I think the more practical solution here is to change it to activate and to go on.
tux [4:49 PM]
I dunno, reconstruct made tons of sense to me. Made it clear that SSSS was happening
If other terms make more sense, use those I guess.
dnunez [4:49 PM]
Yeah, I know...:expressionless:
jmyles [4:50 PM]
@tux: What exactly is being "reconstructed" though?  What was constructed, deconstructed, and constructed again?
tux [4:50 PM]
The secret
jmyles [4:50 PM]
But that never actually happens.
The secret isn't *deconstructed*
KPrasch [4:50 PM]
IMO Designing a metaphor is programmatic labor, when done well, communicates more than useage, but an intuition about how to use the tool. (edited)
tux [4:50 PM]
Yes it does
That's how the threshold split key stuff works
jmyles [4:51 PM]
`Alice` isn't in possession of the exact same `e_prime` that `Bob` will end up with.  She doesn't smash it and await him to build it back together.
...and picturing it that way is confusing.
dnunez [4:53 PM]
The thing is that the reconstructed secret is never actually disclosed since it's always maintained as an exponent
So although we do shamir
We don't do it in the typical way
Bob never gets the shamirs secret
jmyles [4:55 PM]
sure - if he were literally *reconstructing* the same shamir secret, then we'd use a different metaphor.
tux [4:55 PM]
I'd argue that it would still be the same
Reconstruction is still happening
And not in the US History sense
dnunez [4:56 PM]
Yes
dnunez [4:56 PM]
That's why we are not going to solve this issue with a debate about semantics
dnunez [4:57 PM]
I'd prefer to move on
None and both terms are right
jmyles [4:57 PM]
I prefer to get it right, and have a metaphor that we all can really picture and easily explain to people.
KPrasch [4:57 PM]
The difference there is that building an equal object is not that same as building the same (identical) object (edited)
jmyles [4:59 PM]
Consider this line: https://github.com/nucypher/pyUmbral/blob/master/umbral/umbral.py#L203
GitHubGitHub
nucypher/pyUmbral
pyUmbral - Nucypher KMS implementation of Umbral using OpenSSL and Cryptography.io
If this were *reconstruction*, we'd expect not to multiply by `cfrag.point_eph_e1`, but to be able to `assert self._point_eph_e == self._e_prime`.
The `Capsule` never *sheds* anything, to get a new one of that same thing and *reconstruct* anew.  Instead, things only ever area *added* to it.
Listen: if `_reconstruct` is the best thing, let's go with it.  I just don't want to use it only because it's what's there.  I want to use the right word.  This is the essence of our project.
I don't know how to tell the story of `Bob` and the `Capsule` using the word "reconstruct" - it doesn't seem to me that it's the right way to engage people and give them the insight they'll want to realize the gravity of what we're doing.
But this PR took me < 60 seconds to make, so there's really no hard feelings if it's rejected.  :slightly_smiling_face:
dnunez [5:09 PM]
As I said, I can live with this, even if I'm not satisfied. I already accepted the pr

jMyles:
More discussion later:

tux [3:32 PM]
@jmyles I was going over some of the code, then it just occurred to me — just call it `construct`, not `reconstruct`!
It was so obvious
If you don’t agree, I’m still fine with `activate`, but `construct` is pretty damn accurate.
michwill [3:34 PM]
You mean that it was never constructed? Huh
tux [3:35 PM]
Well, it’s like a deconstructed salad
you can construct it
lol
michwill [3:40 PM]
:thinking_face:
That's pretty tricky for me to understand. But whatever
Alice shares capsule with Bob, a construction worker
Doesn't sound better or worse to me than reconstruct, activate.. So all good
tux [3:42 PM]
Yeah, just another perspective
jmyles [4:58 PM]
@tux: I just woke up from a nap with a different idea that you might like: We used the word "activated" to describe the past tense of what has happened after you run `Capsule._reconstruct_shamirs_secret()` (I just updated my PR with this; give a quick look).
(ie, rename `_activate` to `_reconstruct_shamirs_secret()`, but leave `activated_components`, and use the vernacular "activated" to describe a `Capsule` which `Bob` can open)
(possibly also later a method, `is_activated()` if we need it)
So this turns the story into this: You don't reconstruct *the Capsule*, you reconstruct shamir's secret, which in turn *activates* the capsule. (edited)
As for the general wisdom of spending good, wholesome time on naming things and ensuring empowering domain language (including, but not limited to, a coherent system metaphor), I think that's a topic that's worth discussing at the homecoming.  We have perhaps had different experiences, but for my part, I have never looked back on that sort of work and thought, "well, that was a waste."  I have, on the other hand, watched neglected domain language turn into crushing technical debt.  But yeah, let's discuss it in person with the screens put to one side.
I think that, at least for me, a huge part of whether it's time well-spent is whether the dialogue blocks other work or whether it occurs contemporaneously to other progress. (edited)
michwill [5:12 PM]
So what does Bob have to do?
jmyles [5:12 PM]
Bob still uses the higher-level API that @tux designed
But, for example, he can optionally reserialize a `Capsule` and then confirm it with `capsule.activated_components() == ...`
@KPrasch and I had a long talk about `Bob` last night, and some other utilities that he's missing.  For example, currently, how do we expect Bob to keep track of `m` and `n`?
michwill [5:14 PM]
So, Bob does `capsule.attach_cfrag`?
jmyles [5:15 PM]
Also, it might make sense to have a facility wherein `Alice` can seamlessly share both the `Capsule` and the `TreasureMap`, since those two naturally go together for `Bob` to start his journey.
michwill [5:15 PM]
All right, then I think the flow you've made makes total sense
jmyles [5:16 PM]
Right?  This way, there's absolutely no doubt that shamir's is happening, but also no metaphor which seems to describe a shattered `Capsule`
tux [5:16 PM]
Yeah, this is great
KPrasch [5:19 PM]
Yes!

PR #53, opened 2018-02-08
Implement curve agnostic behaviors in operations

tuxxy:
### What this does:
1. Use SHA512 in the cryptographic material hashing functions (`hash_to_bn` and `unsafe_hash_to_point`).
2. Implement `CURVE_KEY_SIZE_BYTES` and `CURVE_MINVAL_SHA512` in `UmbralParameters`.
3. Use `BytesIO` to deserialize fragment objects based on curve keysize.
4. Move `unsafe_hash_to_bn` and `hash_to_bn` to the point and bignum modules, respectively.
    - This solves a circular import issue.
5. Add a `get_curve_keysize_bytes` function to `utils` which accurately returns keysizes for a specific curve.
6. Some reorganization changes.
7. Adds `to_cryptography_priv_key` and `to_cryptography_pub_key` on `BigNum` and `Point`.
8. Adds tests for the above ^.

### DIFF
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -54,6 +54,20 @@ def test_simple_api(N, threshold):
     assert reenc_cleartext == plain_data
 
 
+def test_pub_key_encryption():
+    priv_key_alice = keys.UmbralPrivateKey.gen_key()
+    pub_key_alice = priv_key_alice.get_pub_key()
+
+    priv_key_bob = keys.UmbralPrivateKey.gen_key()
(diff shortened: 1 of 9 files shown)


tests/test_umbral.py
```diff
@@ -54,6 +54,20 @@ def test_simple_api(N, threshold):
     assert reenc_cleartext == plain_data
 
 
+def test_pub_key_encryption():
```
jMyles: Yep, looks groovy.

PR #55, opened 2018-02-09
Point.__mul__ unit and regression test; Pipenv dependency management.

KPrasch:


Pipfile
```diff
@@ -0,0 +1,22 @@
+[[source]]
+
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+
+[packages]
+
+msgpack-python = "*"
```
tuxxy: Remove this pls.

tuxxy: Looks awesome, btw! :D

PR #59, opened 2018-02-09
Mock OpenSSL fixture and Point arithmetic tests.

KPrasch:
Additional test coverage for:
-  Point.__mul__
- Point.__add__
- Point.__invert__
- Point.__eq__


tests/test_point/test_point_arithmetic.py
```diff
@@ -0,0 +1,106 @@
...
+
+import pytest
+from cryptography.hazmat.backends.openssl import backend
+from cryptography.hazmat.primitives.asymmetric import ec
+
+from umbral.bignum import BigNum
+from umbral.config import default_curve
+from umbral.point import Point
+
+curve = default_curve() or ec.SECP256K1()
```
jMyles: Let's just have `CURVE = default_curve()` here.

tests/test_point/test_point_arithmetic.py
```diff
@@ -0,0 +1,106 @@
...
+            assert 'BN_CTX' in str(bn_ctx)
+            assert 'EC_GROUP' in str(group)
+            assert 'EC_POINT' in str(ec_point)
+            assert 'EC_POINT' in str(other_point)
+
+            assert random_ec_point1.group == group
+            assert random_ec_point1.ec_point == ec_point
+            assert random_ec_point1.ec_point == other_point
+
+            return 1
```
jMyles: Perhaps a comment here, explaining that this is the expected output of the mocked function(s)?

tests/test_point/test_point_arithmetic.py
```diff
@@ -0,0 +1,106 @@
...
+            for method, patch in mock_load.items():
+                stack.enter_context(mocker.mock_module.patch.object(backend._lib, method, patch))
+            yield
+
+    return mocked_openssl_backend
+
+
+def test_mocked_openssl_point_arithmetic(mock_openssl, random_ec_point1, random_ec_point2, random_ec_bignum):
+    with mock_openssl():
+        _ = random_ec_point1 == random_ec_point1   # __eq__
```
jMyles: Instead of merely saving and releasing these, why not make assertions about them?

tests/test_point/test_point_arithmetic.py
KPrasch: I ended up removing it completely, and letting the default pass through.

tests/test_point/test_point_arithmetic.py
KPrasch: Possible, but we would have to use the actual openssl methods to generate bool expressions in each case.

KPrasch:
Moved to PR #61

Issue #54, opened 2018-02-08
Implement tests that verify curve agnosticism of Umbral


tuxxy:
Currently, all tests are performed using secp256k1. We should implement a test that uses a curve with a different key size to verify agnosticism.

KPrasch:
Implemented, but the tests do not pass.

tuxxy:
Resolved in #104/#102

PR #61, opened 2018-02-10
Multi-curve and arithmetic test coverage

KPrasch:
Also tests against ec.SECP256K1, ec.SECP384R1, and ec.SECP192R1.


KPrasch:
also fixes issue #54

PR #63, opened 2018-02-11
Umbral test coverage: restores tests to latest umbral api

KPrasch:
Updates test to use latest umbral api; Uncomments an umbral test referenced in issue #6


tests/test_umbral.py
```diff
@@ -8,6 +8,8 @@
 from umbral.params import UmbralParameters
 from umbral.point import Point
 from umbral.umbral import Capsule
+from collections import namedtuple
```
jMyles: Empty?

PR #64, opened 2018-02-12
Umbral Type Hints: return signatures

KPrasch:
PEP 484 Type hints for umbral module;  Focuses on return values.

### DIFF
--- a/tests/test_umbral.py
+++ b/tests/test_umbral.py
@@ -268,46 +268,64 @@ def test_challenge_response_serialization():
     assert new_ch_resp.bn_kfrag_sig2 == ch_resp.bn_kfrag_sig2
     assert new_ch_resp.bn_sig == ch_resp.bn_sig
 
-# @pytest.mark.parametrize("N,M", parameters)
-# def test_cheating_Ursula_replays_old_reencryption(N, M, curve=default_curve()):
-#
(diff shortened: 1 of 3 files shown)


tests/test_umbral.py
```diff
@@ -24,28 +26,40 @@
...
-
+def test_simple_api(alices_keys, bobs_keys, N, M, curve=default_curve()):
     params = UmbralParameters(curve=curve)
-
-    priv_key_alice = keys.UmbralPrivateKey.gen_key(params)
-    pub_key_alice = priv_key_alice.get_pub_key()
-
-    priv_key_bob = keys.UmbralPrivateKey.gen_key(params)
-    pub_key_bob = priv_key_bob.get_pub_key()
+    _, priv_key_alice, pub_key_alice = alices_keys
```
michwill: I know it's from a fixture, but should we have `_something_something` instead of `_`?

KPrasch: OK, good idea.  In this case it's unpacking the keypair's name from the namedtuple
`TestKeyPair = namedtuple('TestKeyPair', 'name priv pub')`.

KPrasch: I ended up removing the TestKeyPair name attr. all together.

PR #65, opened 2018-02-12
[WIP] Check length of bytes on all fragments and Capsule

tuxxy:
### What this does:
1. Adds a utility function `check_bytes_len` which can be used to throw a `ValueError` if the expected and given lengths do not match.
2. Adds this check to `KFrag.from_bytes`, `CapsuleFrag.from_bytes`, and `Capsule.from_bytes`.

### TODO:
1. Add tests.


umbral/fragments.py
```diff
@@ -3,7 +3,7 @@
 from umbral.bignum import BigNum, hash_to_bn
 from umbral.config import default_curve, default_params
 from umbral.point import Point
-from umbral.utils import get_curve_keysize_bytes
+from umbral.utils import get_curve_keysize_bytes, check_bytes_lengtH
```
michwill: lengtH?

PR #66, opened 2018-02-12
GenericUmbralError exception class, and capsule.NotValid.

KPrasch:
Upgrades generic umbral assertions to `raise umbral.GenericUmbralError` or (where appropriate) `raise capsule.NotValid`


KPrasch:
Depends on PR #63 and #64

umbral/umbral.py
```diff
@@ -13,6 +15,10 @@
 from io import BytesIO
 
 
+class GenericUmbralError(Exception):
```
jMyles: Captivating.

PR #76, opened 2018-02-21
Capsule upgrades

jMyles:
Some nice things for `Capsule`:

1. Capsule comparison is now constant time
2. Capsule can now be used as a key in a dict
3. Can't create a Capsule with objects other than Points.

### DIFF
--- a/umbral/umbral.py
+++ b/umbral/umbral.py
@@ -1,3 +1,5 @@
+import hmac
+
 from typing import Tuple, Union, List
 
 from cryptography.hazmat.primitives.asymmetric import ec
@@ -20,6 +22,7 @@ class GenericUmbralError(Exception):
 
 
 class Capsule(object):
+
     def __init__(self,
                  point_eph_e=None,
                  point_eph_v=None,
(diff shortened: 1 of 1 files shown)


umbral/umbral.py
```diff
@@ -129,14 +132,20 @@ def __bytes__(self):
...
+            our_bytes = bytes().join(c.to_bytes() for c in  self.activated_components())
+            other_bytes = bytes().join(c.to_bytes() for c in  other.activated_components())
         elif all(self.original_components() + other.original_components()):
-            original_match = self.original_components() == other.original_components()
-            return original_match
+            our_bytes = bytes().join(c.to_bytes() for c in self.original_components())
+            other_bytes = bytes().join(c.to_bytes() for c in other.original_components())
         else:
             return False
+        return hmac.compare_digest(our_bytes, other_bytes)
```
cygnusv: Why using hmac if it is already provided by cryptography.io?
https://cryptography.io/en/latest/hazmat/primitives/constant-time/#cryptography.hazmat.primitives.constant_time.bytes_eq

tuxxy: Yes, we should definitely use the Cryptography.io one as it is also, probably, based on OpenSSL.

michwill: I think, the reason is because `hmac` is a standard library. But if cryptography.io is better on this front - that's fine

jMyles: Yeah, I think we can go either way - happy to change to the cryptography.io version.  I just thought a sketch with the built-in constant time might be interesting, since we're able to compare bytes.

However, another question dawns on me this morning: since we were previously just comparing lists of elements which are compared in constant time anyway, do we even want to make this change?

It will only ever make a difference if somebody does something truly awful, like using their own `Point` or `BigNum` `class` in `Capsule`.

jMyles:
Superceded by #80.

PR #78, opened 2018-02-23
umbral.umbral -=> umbral.pre; some documentation work also

jMyles:


demo.py
```diff
@@ -1,7 +1,7 @@
 #1
 # Sets a default curve (secp256k1)
 import random
-from umbral import umbral, keys, config
+from umbral import pre, keys, config
```
KPrasch: This namespace is more accurate, IMO. :+1:

umbral/pre.py
```diff
@@ -403,7 +431,7 @@ def decapsulate_reencrypted(pub_key: Point, priv_key: BigNum,
     h = hash_to_bn([e, v], params)
     inv_d = ~d
 
-    if not (s*inv_d) * orig_pub_key == (h*e_prime) + v_prime:
+    if not (s * inv_d) * orig_pub_key == (h * e_prime) + v_prime:
```
KPrasch: PEP8 recommends no spaces here. see: https://www.python.org/dev/peps/pep-0008/#other-recommendations

umbral/pre.py
```diff
@@ -129,15 +138,34 @@ def __bytes__(self):
...
+            # Capsules, determine if a given Capsule is activated.  Do we care about this?
+            # Again, it's hard to imagine why.
             return False
 
+    def __hash__(self):
+        # We only ever want to store in a hash table based on original components;
+        # A Capsule that is part of a dict needs to continue to be lookup-able even
+        # after activation.
+        # Note: In case this isn't obvious, don't use this as a secure hash.  Use BLAKE2b or something.
+        component_bytes = tuple([component.to_bytes() for component in self.original_components()])
```
KPrasch: I notice you are casting the a list comprehension to a tuple, is this because hash value is different if it were a generator expression?  Looks like you can drop the `[ ]`.

demo.py
jMyles: :heavy_check_mark:

PR #85, opened 2018-03-08
Implement methods to convert Umbral keys to cryptography.io keys in `keys` and some other changes.

### LINKED ISSUE #74: Using both Chacha20 from OpenSSL and pynacl. Need to chose one!
In `umbral.py`, we use constants from pynacl, e.g.:
```
    key, capsule = _encapsulate(alice_pubkey.point_key, SecretBox.KEY_SIZE)
```
But for the cipher, we actually use cryptography.io/openssl. In `dem.py`:
```
        self.cipher = ChaCha20Poly1305(symm_key)
```

We need to use consistently either one, or the other

### LINKED ISSUE #84: Shortcut method to get Cryptography.io keys from Umbral keys
Current way is through methods in BigNum and Point classes. But perhaps it has more sense to offer such methods in Umbral keys. We can of course offer them in both places, but it seems more likely to be used in Umbral keys.

Related:
https://github.com/nucypher/nucypher-kms/pull/178/commits/53a0a33f6a68068a579561c9dbd9c4b32b5b001a#diff-f8cdf78f1ab20bfd277c9666b059217eR115

tuxxy:
### What this does:
1. Moves `to_cryptography_pubkey` and `to_cryptography_privkey` to `UmbralPublicKey` and `UmbralPrivateKey` respectively.
2. Adds tests in `test_umbral_keys`.
3. Creates and uses a constant `CHACHA20_KEY_SIZE` for encapsulation instead of the unused `SecretBox.KEY_SIZE` constant.
3. Resolves #84 
4. Resolves #74

### What needs to be done:
1. Update `nucypher-kms` to handle these changes.
2. ~Make an issue with cryptography.io to add key size constants for their specific AEAD and symmetric ciphers.~ [Done](https://github.com/pyca/cryptography/issues/4137)

### DIFF
--- a/umbral/keys.py
+++ b/umbral/keys.py
@@ -209,11 +209,15 @@ class UmbralPublicKey(object):
         backend.openssl_assert(ec_key != backend._ffi.NULL)
         ec_key = backend._ffi.gc(ec_key, backend._lib.EC_KEY_free)
 
-        res = backend._lib.EC_KEY_set_group(ec_key, self.point_key.group)
-        backend.openssl_assert(res == 1)
+        set_group_result = backend._lib.EC_KEY_set_group(
+            ec_key, self.point_key.group
+        )
+        backend.openssl_assert(set_group_result == 1)
 
-        res = backend._lib.EC_KEY_set_public_key(ec_key, self.point_key.ec_point)
-        backend.openssl_assert(res == 1)
+        set_pubkey_result = backend._lib.EC_KEY_set_public_key(
+            ec_key, self.point_key.ec_point
+        )
+        backend.openssl_assert(set_pubkey_result == 1)
 
         evp_pkey = backend._ec_cdata_to_evp_pkey(ec_key)
         return _EllipticCurvePublicKey(backend, ec_key, evp_pkey)
--- a/umbral/keys.py
+++ b/umbral/keys.py
@@ -120,30 +120,34 @@ class UmbralPrivateKey(object):
         backend.openssl_assert(self.bn_key.group != backend._ffi.NULL)
         backend.openssl_assert(self.bn_key.bignum != backend._ffi.NULL)
 
-        new_ec_key = backend._lib.EC_KEY_new()
-        backend.openssl_assert(new_ec_key != backend._ffi.NULL)
-        ec_key = backend._ffi.gc(new_ec_key, backend._lib.EC_KEY_free)
+        ec_key = backend._lib.EC_KEY_new()
+        backend.openssl_assert(ec_key != backend._ffi.NULL)
+        ec_key = backend._ffi.gc(ec_key, backend._lib.EC_KEY_free)
 
-        set_group_result = backend._lib.EC_KEY_set_group(ec_key, self.bn_key.group)
+        set_group_result = backend._lib.EC_KEY_set_group(
+            ec_key, self.bn_key.group
+        )
         backend.openssl_assert(set_group_result == 1)
 
-        set_private_key_result = backend._lib.EC_KEY_set_private_key(ec_key, self.bn_key.bignum)
-        backend.openssl_assert(set_private_key_result == 1)
+        set_privkey_result = backend._lib.EC_KEY_set_private_key(
+            ec_key, self.bn_key.bignum
+        )
+        backend.openssl_assert(set_privkey_result == 1)
 
         # Get public key
-        new_point = backend._lib.EC_POINT_new(self.bn_key.group)
-        backend.openssl_assert(new_point != backend._ffi.NULL)
-        point = backend._ffi.gc(new_point, backend._lib.EC_POINT_free)
+        point = backend._lib.EC_POINT_new(self.bn_key.group)
+        backend.openssl_assert(point != backend._ffi.NULL)
+        point = backend._ffi.gc(point, backend._lib.EC_POINT_free)
 
         with backend._tmp_bn_ctx() as bn_ctx:
-            multiplication_result = backend._lib.EC_POINT_mul(
+            mult_result = backend._lib.EC_POINT_mul(
                 self.bn_key.group, point, self.bn_key.bignum, backend._ffi.NULL,
                 backend._ffi.NULL, bn_ctx
             )
-        backend.openssl_assert(multiplication_result == 1)
+            backend.openssl_assert(mult_result == 1)
 
-        set_public_key_result = backend._lib.EC_KEY_set_public_key(ec_key, point)
-        backend.openssl_assert(set_public_key_result == 1)
+        set_pubkey_result = backend._lib.EC_KEY_set_public_key(ec_key, point)
+        backend.openssl_assert(set_pubkey_result == 1)
 
         evp_pkey = backend._ec_cdata_to_evp_pkey(ec_key)
         return _EllipticCurvePrivateKey(backend, ec_key, evp_pkey)
--- a/umbral/keys.py
+++ b/umbral/keys.py
@@ -3,10 +3,13 @@ import base64
 
 from typing import Union
 
+from nacl.secret import SecretBox
 from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.kdf.scrypt import Scrypt
 from cryptography.hazmat.backends import default_backend
-from nacl.secret import SecretBox
+from cryptography.hazmat.backends.openssl.ec import (
+    _EllipticCurvePublicKey, _EllipticCurvePrivateKey
+)
 
 from umbral.config import default_params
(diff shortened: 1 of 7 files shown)


tuxxy:
@jMyles I don't see a point in putting this logic in `Point` and `BigNum` anymore. Furthermore, @cygnusv and I are starting to look at that more as a primitive prototyping interface and we'll probably move that out eventually to its own library.

Essentially, this logic would have eventually ended up in `UmbralPublicKey` and `UmbralPrivateKey` anyway. There is no repeating logic, either. This has completely moved it to it.

jMyles:
Groovy, that sounds great.

> @cygnusv and I are starting to look at that more as a primitive prototyping interface and we'll probably move that out eventually to its own library.

Did I miss a memo?  I want to be part of that convo.

tuxxy:
Nah, it's just in the Slack somewhere. You didn't miss anything.

umbral/keys.py
```diff
@@ -108,6 +111,43 @@ def get_pubkey(self):
...
+        Returns a cryptography.io EllipticCurvePrivateKey from the Umbral key.
+        """
+        backend = default_backend()
+
+        backend.openssl_assert(self.bn_key.group != backend._ffi.NULL)
+        backend.openssl_assert(self.bn_key.bignum != backend._ffi.NULL)
+
+        ec_key = backend._lib.EC_KEY_new()
+        backend.openssl_assert(ec_key != backend._ffi.NULL)
+        ec_key = backend._ffi.gc(ec_key, backend._lib.EC_KEY_free)
```
jMyles: This pattern is unclean IMO.

Why do we name `ec_key`, only to shadow it two lines later with a different object?  (I notice this happens on two other occasions in this PR).

It reminds me of the similar function in cryptography.io, `_evp_pkey_to_public_key`.  That always bothered me there too.

Can't we do better?

umbral/keys.py
```diff
@@ -108,6 +111,43 @@ def get_pubkey(self):
...
+
+        res = backend._lib.EC_KEY_set_group(ec_key, self.bn_key.group)
```
jMyles: I don't think I like `res` as the name of the result of every call to the backend.  Can't we have more informative and unique names?

In addition to being confusing, it's a bit dangerous: if I comment out one of the intermediate lines, the assertion still passes, right?  Don't we not want that?

umbral/keys.py
```diff
@@ -108,6 +111,43 @@ def get_pubkey(self):
...
+        res = backend._lib.EC_KEY_set_private_key(ec_key, self.bn_key.bignum)
+        backend.openssl_assert(res == 1)
+
+        # Get public key
+        point = backend._lib.EC_POINT_new(self.bn_key.group)
+        backend.openssl_assert(point != backend._ffi.NULL)
+        point = backend._ffi.gc(point, backend._lib.EC_POINT_free)
+
+        with backend._tmp_bn_ctx() as bn_ctx:
+            res = backend._lib.EC_POINT_mul(
```
jMyles: We do point multiplication here with `NULL` - although I have no doubt this is correct, it's not obvious why it's correct.  What's actually happening here?

umbral/keys.py
```diff
@@ -152,6 +192,28 @@ def to_bytes(self):
...
+        Returns a cryptography.io EllipticCurvePublicKey from the Umbral key.
+        """
+        backend = default_backend()
+
+        backend.openssl_assert(self.point_key.group != backend._ffi.NULL)
+        backend.openssl_assert(self.point_key.ec_point != backend._ffi.NULL)
+
+        ec_key = backend._lib.EC_KEY_new()
+        backend.openssl_assert(ec_key != backend._ffi.NULL)
+        ec_key = backend._ffi.gc(ec_key, backend._lib.EC_KEY_free)
```
jMyles: Is it not possible / practical to do `_lib.EVP_PKEY_free` here instead?

umbral/keys.py
tuxxy: No, this is how OpenSSL works. :P

umbral/keys.py
tuxxy: It's a `res` as much as a `result`. It either outputs `1`, `0`, or in some cases `-1`.

umbral/keys.py
tuxxy: Why would I do that? It's an EC_KEY?

umbral/keys.py
tuxxy: From the OpenSSL docs on EC_POINT_mul, "EC_POINT_mul calculates the value generator * n + q * m and stores the result in r. The value n may be NULL in which case the result is just q * m."

So, it's only used to perform a multiplication without the other stuff.

umbral/keys.py
jMyles: Yyyyeahhh, I think I was kinda wilin' there.  It was late.  :-)

I was imagining a flow like:

```
pkey = _lib.EVP_PKEY_new()
self._pkey = _ffi.gc(pkey, _lib.EVP_PKEY_free)
backend._lib.EC_KEY_new()
...
backend._lib.EVP_PKEY_assign_EC_KEY(...)
```

But yeah, I see now that that does not make sense.

umbral/keys.py
jMyles: Ahh, got it.  Nice.

umbral/keys.py
jMyles: Yeah, I mean I know that.  But what you do mean it's a "res as much as a result"?  As I asked above, why got have informative and unique names?  That's what I always do.

umbral/keys.py
jMyles: Heh, well it doesn't have to be.  Why not change them?

tuxxy: Well, an EVP_PKEY is a structure for holding a private key. An EC_KEY essentially like the primitive building block of it, so an EVP_PKEY still requires an EC_KEY.

jMyles: No no, I know that.  I'm talking about the shadowed names.

umbral/keys.py
tuxxy: I mean you could change `res` to `ret` or `ret_code`, but pretty much anyone who looks at this code is going to understand what `res` is.

jMyles: Sure, but those are equally dumb names.  I'm talking about actual unique names.

umbral/keys.py
```diff
@@ -108,6 +111,43 @@ def get_pubkey(self):
...
+        backend.openssl_assert(res == 1)
+
+        # Get public key
+        point = backend._lib.EC_POINT_new(self.bn_key.group)
+        backend.openssl_assert(point != backend._ffi.NULL)
+        point = backend._ffi.gc(point, backend._lib.EC_POINT_free)
+
+        with backend._tmp_bn_ctx() as bn_ctx:
+            res = backend._lib.EC_POINT_mul(
+                self.bn_key.group, point, self.bn_key.bignum, backend._ffi.NULL,
```
jMyles: Why not close the context manager here?  You don't need the OpenSSL context for any operations other than the multiplication, right?

tuxxy: I think so, but IIRC this is a pattern with OpenSSL FFI binding dev. You keep all the used points within the context.

jMyles: Eh?  Really?  It seems sloppy to me.  Is there a doc that explains the advantages of this?

umbral/keys.py
michwill: `success_code`?

jMyles: I mean like this:

```
        new_ec_key = backend._lib.EC_KEY_new()
        backend.openssl_assert(new_ec_key != backend._ffi.NULL)
        ec_key = backend._ffi.gc(new_ec_key, backend._lib.EC_KEY_free)

        set_group_result = backend._lib.EC_KEY_set_group(ec_key, self.bn_key.group)
        backend.openssl_assert(set_group_result == 1)
```

michwill: Ah.. Makes sense

jMyles: Here's a PR: https://github.com/tuxxy/pyUmbral/pull/1

umbral/keys.py
jMyles: @michwill, @tuxxy, so just to close the loop on the gc discussion, `del` does indeed work (at least for the purposes of the tests):

```
new_ec_key = backend._lib.EC_KEY_new()
backend.openssl_assert(new_ec_key != backend._ffi.NULL)
ec_key = backend._ffi.gc(new_ec_key, backend._lib.EC_KEY_free)
del new_ec_key
```

And while I won't persist in this argument if y'all just want to keep `ec_key` for both, I do think that I like this pattern (and the clarity it provides) better.

PR #88, opened 2018-03-13
Deweaponize test data language

KPrasch:
I know that "attack at dawn" is part of a traditional cryptological narrative, but paired with "launch code 0000", I find this... a bit militant. :-) I believe we can communicate good intent by remaining vigilant in our language choices, here in tests, where example stories are told.


michwill:
True. Better to promote thinking that encryption is used for good

tuxxy:
@KPrasch Gave me the go ahead

PR #91, opened 2018-03-18
Bob vs Ursula

tpmccallum:
Does Bob or Usrula attach the cfrags to the capsule? 
I would think that Ursula would as she is already beavering away with the following function

```

cfrag = umbral.reencrypt(kfrag, capsule)

```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,7 +53,7 @@ extending the traditional cryptological narrative of "Alice and Bob" by introduc
 .. code-block:: python
 
   # Ursula re-encrypts the capsule to obtain a cfrag.
-  # Bob attaches the cfrags to the capsule.
+  # Ursula attaches the cfrags to the capsule.
   for kfrag in kfrags:
       cfrag = umbral.reencrypt(kfrag, capsule)
       capsule.attach_cfrag(cfrag)


michwill:
The idea is that to get a capsule which Bob can decrypt, he combines `cfrags` produces by multiple Ursulas from the same initial `capsule`. Bob needs at least `m` out of `n` `cfrags` for that

michwill:
And it's Bob who "attaches" `cfrags` once he got them from `m` Ursulas

cygnusv:
I was currently going over the example in the README and also think that those lines are a little bit unfortunate. Since the example is very simple, it's difficult to understand clearly our intentions. @michwill comment is right, it's Bob who attaches the cfrags to his local copy of the capsule. We have to find a better way to illustrate that in the example.

tpmccallum:
Hi,
Thanks, from what I understand, this really lends itself to decentralization (especially the many Ursula's part). Can you guys please have a look at this GitHub page [1] which is trying to distill how each component/function of pyUmbral will work in conjunction with Ethereum; more specifically Vyper smart contract[s]. I am super keen to build a prototype soon (this week if possible).
Have you built any parts of this inside smart contracts as yet?
I am thinking about modeling this "information selling" against Ethereum's recently ported "safe remote purchases" contract [2].
Any and all feedback is welcome.
Many thanks
Tim

[1] https://github.com/CyberMiles/tim-research/blob/master/information_purchasing/information_purchasing.md
[2] https://github.com/CyberMiles/tim-research/blob/master/information_purchasing/information_purchasing.md

PR #92, opened 2018-03-18
Improvements to reconstruction and challenge protocol. 

### LINKED ISSUE #60: Use BLAKE2b instead of SHA512
Implement BLAKE2b for the KDF instead of SHA512. Pretty straight forward.

### LINKED ISSUE #87: Some tests create re-encryption keys from Alice to Alice
Some tests do something weird, IMHO. Instead of using Alice and Bob as delegator and delegatee, they use Alice for both. For example:
```
def test_challenge_response_serialization():
    priv_key = pre.gen_priv()

    _unused_key, capsule = pre._encapsulate(pub_key)
    kfrags = pre.split_rekey(priv_key, pub_key, 1, 2)
```

Is this intentional? If so, what's the rationale for that? 
I'm currently preparing a PR with some improvements, so I can change this if necessary.

### LINKED ISSUE #90: Drop pre.gen_priv() and pre.priv2pub()?
We currently have two different ways of producing public and private keys in Umbral:

- Through the PRE class:

```
    priv_key = pre.gen_priv()
```
- Through the UmbralPrivateKey class:
```
    priv_key_alice = keys.UmbralPrivateKey.gen_key()
    pub_key_alice = priv_key_alice.get_pubkey()
```
I'd prefer to have just one, preferably the second. What do you think?
As mentioned in my last issue, I'm preparing a small PR so I can make this change if you agree.

cygnusv:
# What this does:

- KFrag ids can only be computed by Bob now
- Added metadata as optional input to the challenge protocol
- Changed SHA512 by BLAKE2b (Resolves #60)
- Dropped pre.gen_priv() and pre.priv2pub() (Resolves #90)
- Fixed some undesired behaviors in tests (Resolves #87)
- Fixed notation, variable names, method visibility, etc
- Some fixes to the README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,13 +63,6 @@ extending the traditional cryptological narrative of "Alice and Bob" by introduc
                           ciphertext, alices_public_key)
 
 
-Features
-==========
-- Proxy Re-encryption Toolkit
-- Re-encryption Key Fragmentation
-- Elliptic Curve Arithmetic
-
-
 Quick Installation
 ==================
 
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,7 @@ extending the traditional cryptological narrative of "Alice and Bob" by introduc
 .. _OpenSSL: https://www.openssl.org/
 
 
-**Encapsulation**
+**Encryption**
 
 .. code-block:: python
 
@@ -45,7 +45,7 @@ extending the traditional cryptological narrative of "Alice and Bob" by introduc
     bobs_public_key = private_key.get_pubkey()
 
     # Alice generates split re-encryption keys for Bob with "M of N".
-    kfrags, _ = umbral.split_rekey(alices_private_key, bobs_public_key, 10, 20)
+    kfrags = umbral.split_rekey(alices_private_key, bobs_public_key, 10, 20)
 
 
 **Re-encryption**
@@ -94,7 +94,7 @@ Academic Whitepaper
 The Umbral scheme academic whitepaper and cryptographic specifications
 are availible on GitHub_.
 
-  "Umbral A Threshold Proxy Re-Encryption Scheme"
+  "Umbral: A Threshold Proxy Re-Encryption Scheme"
   *by David Nuñez*
   https://github.com/nucypher/umbral-doc/blob/master/umbral-doc.pdf
 
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ pyUmbral
 .. image:: https://travis-ci.org/nucypher/pyUmbral.svg?branch=master
     :target: https://travis-ci.org/nucypher/pyUmbral
 
-pyUmbral is a python implementation of David Nuñez's threshold proxy rencryption scheme: Umbral_.
+pyUmbral is a python implementation of David Nuñez's threshold proxy re-encryption scheme: Umbral_.
 Implemented with OpenSSL_ and Cryptography.io_, pyUmbral is a referential and open-source cryptography library
 extending the traditional cryptological narrative of "Alice and Bob" by introducing a new actor,
 *Ursula*, who has the ability to take secrets encrypted for Alice and *re-encrypt* them for Bob.
@@ -65,9 +65,8 @@ extending the traditional cryptological narrative of "Alice and Bob" by introduc
 
 Features
 ==========
-- Re-encryption Toolkit
+- Proxy Re-encryption Toolkit
 - Re-encryption Key Fragmentation
-- Key Encapsulation
 - Elliptic Curve Arithmetic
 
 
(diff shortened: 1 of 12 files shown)


README.rst
```diff
@@ -18,7 +18,7 @@ extending the traditional cryptological narrative of "Alice and Bob" by introduc
 .. _OpenSSL: https://www.openssl.org/
 
 
-**Encapsulation**
+**Encryption**
```
KPrasch: Not to be confused, with Encapsulation.

README.rst
```diff
@@ -65,9 +65,8 @@ extending the traditional cryptological narrative of "Alice and Bob" by introduc
 
 Features
 ==========
-- Re-encryption Toolkit
+- Proxy Re-encryption Toolkit
```
KPrasch: I pondered including "Proxy" here, but since there is no network proxy tooling anywhere in pyUmbral (it's in the KMS), I omitted it to not be misleading; Nonetheless - your intention here is to indicate Umbral's *cryptographic*, not network notion of a proxy?

cygnusv: OK I understand your intention, but since this is very decoupled from the KMS, I think there is not much risk of confusion.

README.rst
cygnusv: Yeah, basically the whole "Encapsulation/Decapsulation" concepts should be only internal to pyUmbral. I think this is clearer now, since the PRE module doesn't expose publicly encapsulation/decapsulation functions anymore (these changes are also included in this PR)

README.rst
KPrasch: I don't know If I agree that there is not much risk of confusion (at least when examining pyUmbral from the perspective of trying understand the Umbral scheme and NuCypher holistically) since a network proxy is such an essential component of most obvious use cases. Perhaps any confusion people experience today is a side-effect of not having robust KMS docs published yet.

README.rst
jMyles: I'm not sure I fully agree with the assertion that "PRE module doesn't expose publicly encapsulation/decapsulation functions".  To my way of thinking, all of the methods on `Capsule` are "encapsulation/decapsulation functions".  I don't think it makes sense to have a public object called a `Capsule` without also treating its narrative as a public one.

README.rst
jMyles: Hmm, I think I'm with @KPrasch on this one.  The toolkit in question doesn't do anything that's particular to a *proxy*, right?  As opposed to re-encryption performed by some other agent?

Or is it customary to *always* regard anyone perform re-encryption as a "proxy"?

tests/test_capsule/test_capsule_operations.py
```diff
@@ -85,39 +85,11 @@ def test_capsule_as_dict_key(alices_keys):
     capsule.attach_cfrag(cfrag)
 
     # Even if we activate the capsule, it still serves as the same key.
-    cleartext = pre.decrypt(capsule, alices_keys.priv, ciphertext, alices_keys.pub)
+    cleartext = pre.decrypt(ciphertext, capsule, alices_keys.priv, alices_keys.pub)
```
jMyles: leaky-chimney

tests/test_challenge.py
```diff
@@ -6,16 +6,21 @@
...
+
+    _unused_key, capsule = pre._encapsulate(pub_key_alice.point_key)
+    kfrags = pre.split_rekey(priv_key_alice, pub_key_bob, 1, 2)
 
     cfrag = pre.reencrypt(kfrags[0], capsule)
 
     capsule.attach_cfrag(cfrag)
-    ch_resp = pre.challenge(kfrags[0], capsule, cfrag)
+
+    metadata = b"Challenge metadata"
```
jMyles: It's not obvious to me what this metadata is doing.  It is supposed to be?

tests/test_challenge.py
```diff
@@ -44,47 +49,55 @@ def test_cheating_ursula_replays_old_reencryption(N, M):
...
+                                   pub_key_alice.point_key,
+                                   pub_key_bob.point_key,
+                                   metadata
+                                   )
+
+    # The response of cheating Ursula is in capsules[0],
+    # so the rest of challenges chould be correct:
+    for i, challenge in enumerate(challenges[1:], 1):
+        cfrag = cfrags[i]
+        metadata = ("Challenge metadata: index {}".format(i)).encode()
```
jMyles: Now I'm even less clear re: metadata.

tests/test_challenge.py
cygnusv: This metadata adds support to include additional information as part of the input from the challenger in the challenger protocol, for example, in order to avoid replays by including a timestamp.

PR #93, opened 2018-03-19
Use BN_rand_range to generate random BigNums within the order of the curve

### LINKED ISSUE #86: Use OpenSSL BN_rand functions instead of Python level functions
Currently we generate random numbers via urandom and check to see if they're a valid private key.

We can use the OpenSSL `BN_rand` functions to do this instead. I've exposed these in cryptography.io: https://github.com/pyca/cryptography/pull/4118

tuxxy:
### What this does:
1. Uses OpenSSL's `BN_rand_range` functions to generate a random BigNum within the order of the curve.
2. Updates to cryptography.io 2.2
3. Fixes #86

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,7 +7,7 @@ name = "pypi"
 
 [packages]
 
-cryptography = "*"
+cryptography = "==2.2"
 pynacl = "*"
 
 
@@ -22,3 +22,4 @@ pytest-mock = "*"
 codecov = "*"
 sphinx = "*"
 sphinx-autobuild = "*"
+cryptography = "==2.2"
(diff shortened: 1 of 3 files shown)


Pipfile
```diff
@@ -7,7 +7,7 @@ name = "pypi"
 
 [packages]
 
-cryptography = "*"
+cryptography = "==2.2"
```
michwill: Idea:
what if we say `">=2.2"`?

tuxxy: Well, we can do `"*"`. I just need to confirm from @KPrasch or @jMyles that this will indeed use the latest version.

michwill: Technically, >=2.2 is probably the correct one if you use the new API

PR #96, opened 2018-03-21
Add GPLv3 license, this software is Free as in Freedom.

tuxxy:
The title speaks for itself. :)

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,675 @@
+### GNU GENERAL PUBLIC LICENSE
+
+Version 3, 29 June 2007
+
+Copyright (C) 2007 Free Software Foundation, Inc.
+<https://fsf.org/>
+
+Everyone is permitted to copy and distribute verbatim copies of this
+license document, but changing it is not allowed.
+
+### Preamble
+
+The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works. By contrast,
+the GNU General Public License is intended to guarantee your freedom
+to share and change all versions of a program--to make sure it remains
+free software for all its users. We, the Free Software Foundation, use
+the GNU General Public License for most of our software; it applies
+also to any other work released this way by its authors. You can apply
+it to your programs, too.
+
+When we speak of free software, we are referring to freedom, not
+price. Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
+
+To protect your rights, we need to prevent others from denying you
+these rights or asking you to surrender the rights. Therefore, you
+have certain responsibilities if you distribute copies of the
+software, or if you modify it: responsibilities to respect the freedom
+of others.
+
+For example, if you distribute copies of such a program, whether
+gratis or for a fee, you must pass on to the recipients the same
+freedoms that you received. You must make sure that they, too, receive
+or can get the source code. And you must show them these terms so they
+know their rights.
+
+Developers that use the GNU GPL protect your rights with two steps:
+(1) assert copyright on the software, and (2) offer you this License
+giving you legal permission to copy, distribute and/or modify it.
+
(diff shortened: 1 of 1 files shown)


jMyles:
Another question on my mind at this later hour of licensing talk:

In the past we've talked about releasing a "client" variant of the KMS (ie, with `Alice` and `Bob`, but not necessarily `Ursula`) for use in proprietary projects.  If this variant uses pyumbral, that will be off the table.  Are we OK with that?

michwill:
I think so. We can change the license to, say, LGPL, if we find it being a problem. Also, we can have javascript client being under a permissive license

tuxxy:
Why would someone make a proprietary client? I am against all such things
being made for the public KMS network for the safety of our users.

michwill:
Some corporate customers may want closed-source solutions. But they'd probably like to license in such case

tuxxy:
Corporate customers don't need to distribute their source code, though. The
GPL still allows them to use it.

michwill:
So.. We're good to merge?

tuxxy:
I think so, waiting for approvals

jMyles:
> Why would someone make a proprietary client? I am against all such things being made for the public KMS network for the safety of our users.

This is an odd argument.  Of course someone can ignore the license and do it anyway, as we have discussed.  The "safety of our users" *does not* and *can not* depend on people being honest vis a vis the license.  So let's be clear about that.

So, @tuxxy, @michwill, just to be clear: none of the current use-cases (like the medical records folks, for example) are wanting to use the KMS codebase (or at least an Alice-Bob variant) in a proprietary setting?  Or, if they are, we're going to tell them they can't (and hope they oblige)?

> Corporate customers don't need to distribute their source code, though. The GPL still allows them to use it.

Wait, this again?  What exactly do you mean?  It's totally possible (in fact, as I understand it, part of the plan) for some of the known userbase to distribute their *program* which uses the KMS codebase (and thus pyumbral).  If they do that, they'll also need to supply their customers with their entire codebase.

tuxxy:
Can you explain to me how you can easily distribute a closed source Python
program?

michwill:
I think, @tuxxy may be referring to SaaS products (or interaction with a local daemon over sockets), since we're not using AGPL

michwill:
One can obfuscate and distribute `*pyc`, perhaps

michwill:
Here are the methods https://wiki.python.org/moin/Asking%20for%20Help/How%20do%20you%20protect%20Python%20source%20code%3F

Out of all those, GPL allows SaaS

jMyles:
@tuxxy: Yeah, I mean it's not as easy as with C obviously, but people usually either distribute the bytecode (as @michwill points out) or use Cython or something.

But in this case, they needn't even do that: what if they sell a proprietary device which uses the code in question?  That's enough to trigger GPL.  Or do we presume they'll rent or something?

michwill:
Actually, in the case of selling proprietary devices it is exactly the case when a commercial license for them will do (e.g. typical GPL + commercial scheme)

jMyles:
I mean, look, maybe we just tell them that, if they want to sell such devices to people, they have to include a copy of their codebase to that person. They can still ask the user to sign a ToS saying they won't further distribute it.

At the end of the day, that's the point, right? If you're going to build a medical device using our KMS, you need to tell your user exactly what you're doing. I can get behind that, but I just want to be clear about the implications.

They can also still rent their devices out, but that's a less savory solution IMO - I almost wish GPL *didn't* allow that.  (Or maybe it doesn't?)

jMyles:
@michwill: So, in that scenario, NuCypher will retain ownership and license the codebase on a case-by-case commercial basis?

michwill:
Yes.
But really, in many hardware cases it isn't even required to do it. I've seen plenty of routers combining GPL software with proprietary (where GPL was even printed out)

jMyles:
@michwill: Right, but my understanding is that companies who make routers which use GPL do indeed need to provide the entire source code to their customers (as linksys was famously forced to do).

michwill:
Hm. So they cannot have Linux (under GPL) and some proprietary web UI running in it?

jMyles:
My understanding is that this has never made it to court - the closest was the linksys case, and linksys backed down before it got to court.

PR #102, opened 2018-03-31
Use params injection on necessary functions

### LINKED ISSUE #54: Implement tests that verify curve agnosticism of Umbral

### LINKED ISSUE #99: Consistent use of UmbralParameters in the main PRE API
Currently, `UmbralParameters` is an input parameter of `split_rekey()` and `reencrypt()`, but not in `encrypt()` and `decrypt()`. Strangely, it is a parameter of internal functions like `_encapsulate()`, `_decapsulate_original()` and `_decapsulate_reencrypted()`.

I propose to remove it from the internal functions, and put it only in the public facing ones.

tuxxy:
### What this does:
1. Adds `params` arg to a few calls, primarily `encrypt`, `decrypt`, `_open_capsule`, `_encapsulate`, `_decapsulate_original`, `_decapsulate_reencrypted`, and `_reconstruct_shamirs_secret`. (resolves #99)
2. Updates API for tests for use on multiple curves. (resolves #54)

### DIFF
--- a/tests/test_capsule/test_capsule_operations.py
+++ b/tests/test_capsule/test_capsule_operations.py
@@ -85,7 +85,7 @@ def test_capsule_as_dict_key(alices_keys):
     capsule.attach_cfrag(cfrag)
 
     # Even if we activate the capsule, it still serves as the same key.
-    cleartext = pre.decrypt(ciphertext, capsule, alices_keys.priv, alices_keys.pub)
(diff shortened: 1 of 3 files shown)


tuxxy:
Merged in #104

PR #105, opened 2018-04-03
use abstract class to gain type safety

kikofernandez:
Hi,

I looked through the code and I thought that there was a repeated pattern.

You could use an abstract class that defines the methods `from_bytes`, `to_bytes` and `__bytes__`, and make `from_bytes` and `to_bytes` abstract methods. This means that any derived class (of the abstract class) needs to implement the abstract methods and if not, you get immediately a runtime error. This is useful as a way to ensure that your refactoring doesn't delete those methods by mistake

In this code, the abstract class contains a method `__bytes__` that is not always used in all the classes but, as far as I can see, it should present no problems. 

I could not run the code to test it but I hope that the main idea is understood.

P.S. I have removed imports that PyCharm claims to not be used.

### DIFF
--- a/umbral/keys.py
+++ b/umbral/keys.py
@@ -1,10 +1,7 @@
 import os
 import base64
 
-from typing import Union
-
 from nacl.secret import SecretBox
-from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.kdf.scrypt import Scrypt
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.backends.openssl.ec import (
@@ -14,9 +11,10 @@ from cryptography.hazmat.backends.openssl.ec import (
 from umbral.config import default_params
 from umbral.point import Point, BigNum
 from umbral.params import UmbralParameters
+from umbral.utils import AbstractCryptoEntity
 
 
-class UmbralPrivateKey(object):
+class UmbralPrivateKey(AbstractCryptoEntity):
     def __init__(self, bn_key: BigNum, params: UmbralParameters=None):
         """
         Initializes an Umbral private key.
--- a/umbral/utils.py
+++ b/umbral/utils.py
@@ -1,7 +1,9 @@
 import math
+from abc import ABC, abstractmethod
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
 
 
@@ -42,3 +44,23 @@ def kdf(ecpoint, key_length):
 
 def get_curve_keysize_bytes(curve):
     return int(math.ceil(curve.key_size / 8.00))
+
+
+class AbstractCryptoEntity(ABC):
+    @classmethod
+    @abstractmethod
+    def from_bytes(cls, *args, **kwargs):
+        """
+        Instantiate a split-key fragment object from the serialised data.
+        """
+        raise NotImplementedError("Derived classes should implement this method")
+
+    @abstractmethod
+    def to_bytes(self):
+        """
+        Serialize a split-key into a bytestring.
+        """
+        raise NotImplementedError("Derived classes should implement this method")
+
+    def __bytes__(self):
+        return self.to_bytes()
--- a/umbral/fragments.py
+++ b/umbral/fragments.py
@@ -3,12 +3,13 @@ from cryptography.hazmat.primitives.asymmetric import ec
 from umbral.bignum import BigNum, hash_to_bn
 from umbral.config import default_curve, default_params
 from umbral.point import Point
-from umbral.utils import get_curve_keysize_bytes
+from umbral.utils import get_curve_keysize_bytes, AbstractCryptoEntity
+
 
 from io import BytesIO
 
 
-class KFrag(object):
+class KFrag(AbstractCryptoEntity):
     def __init__(self, id_, key, x, u1, z1, z2):
         self.bn_id = id_
         self.bn_key = key
@@ -17,8 +18,8 @@ class KFrag(object):
         self.bn_sig1 = z1
         self.bn_sig2 = z2
 
-    @staticmethod
-    def from_bytes(data: bytes, curve: ec.EllipticCurve = None):
+    @classmethod
+    def from_bytes(cls, data: bytes, curve: ec.EllipticCurve = None):
         """
         Instantiate a KFrag object from the serialized data.
         """
@@ -35,7 +36,7 @@ class KFrag(object):
         sig1 = BigNum.from_bytes(data.read(key_size), curve)
         sig2 = BigNum.from_bytes(data.read(key_size), curve)
 
-        return KFrag(id, key, eph_ni, commitment, sig1, sig2)
+        return cls(id, key, eph_ni, commitment, sig1, sig2)
 
     def to_bytes(self):
         """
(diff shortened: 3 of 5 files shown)


tuxxy:
Hey @kikofernandez thanks for the PR.

It's my understanding that the team thinks using an abstract base class for Bytestring serialization is a bit heavy handed here right now. We have our own vision for how this all moves forward and we have some ideas on that.

kikofernandez:
Got it.

kikofernandez:
Can I suggest that someone creates a `CONTRIBUTING.d` file to know other people can contribute? Your response was really polite but it does not let me or others know what kind of help is appreciated / needed and if we should create an issue and explain a possible PR before actually submitting the PR.

Thanks.

KPrasch:
Hmm, I think such a change is a step in maturing the codebase (keeping maintainability, high test coverage, and guided extensibility); I'd like to see this implemented.  If we use an abstract base inheritance scheme, we can implement `@abstractmethod`/ `NotImplemented` as @kikofernandez  suggests for serialization. :+1:

KPrasch:
@kikofernandez, We're authoring official NuCypher KMS documentation presently, and we've also recently licensed pyUmbral! Nonetheless, we do want to include contribution guidelines for pyUmbral in the near future.

umbral/utils.py
```diff
@@ -42,3 +44,23 @@ def kdf(ecpoint, key_length):
...
 def get_curve_keysize_bytes(curve):
     return int(math.ceil(curve.key_size / 8.00))
+
+
+class AbstractFromToBytes(ABC):
+    @staticmethod
+    @abstractmethod
+    def from_bytes(cls, data: bytes, curve: ec.EllipticCurve = None):
+        """
+        Instantiate a KFrag object from the serialized data.
```
cygnusv: The comments on this file should be generic

umbral/utils.py
```diff
@@ -42,3 +44,23 @@ def kdf(ecpoint, key_length):
 
 def get_curve_keysize_bytes(curve):
     return int(math.ceil(curve.key_size / 8.00))
+
+
+class AbstractFromToBytes(ABC):
+    @staticmethod
+    @abstractmethod
+    def from_bytes(cls, data: bytes, curve: ec.EllipticCurve = None):
```
cygnusv: I have some problems with this part... The `curve` parameter only makes sense in some classes (e.g, BigNum, Point), but not in others (e.g., KFrag). Being an optional parameter makes this problem not very relevant, but at the same time it means we have to explicitly throw an error or something similar when it is used incorrectly (e.g., inputing a `curve` when deserializing a KFrag).

umbral/keys.py
```diff
@@ -1,10 +1,7 @@
 import os
 import base64
 
-from typing import Union
-
 from nacl.secret import SecretBox
-from cryptography.hazmat.primitives.asymmetric import ec
```
cygnusv: Thanks for the housekeeping!

kikofernandez: no problem :)

umbral/utils.py
kikofernandez: well, what if we abstract away all the arguments? Then each derived class should know what to do, since this is not an implementation, just a way to force an implementation, e.g.

```python
class AbstractFromTo(ABC):
  @staticmethod
  @abstractmethod
  def from_bytes(*args, **kwargs):
    pass
```
and the same with the other methods

umbral/utils.py
kikofernandez: Would it work as `Instantiate a cls object from the serialised data.`?

umbral/utils.py
kikofernandez: BTW, I just saw that the `Fragments` breaks the `from_bytes` format used in the other classes, [code here](https://github.com/nucypher/pyUmbral/pull/105/files/bc586efe57fc80a9c472224d00d59fd109562afe#diff-bce02964095b01c806e4cc506eda1ee0L21):

```python
class KFrag(AbstractFromToBytes):
  ...
    @staticmethod
    def from_bytes(data: bytes, curve: ec.EllipticCurve = None):
        pass
```
while all other versions use a `@classmethod`. 
Do you want to keep consistency `from_bytes` and `to_bytes` or it was written as such for a specific reason?

umbral/utils.py
```diff
@@ -42,3 +44,23 @@ def kdf(ecpoint, key_length):
 
 def get_curve_keysize_bytes(curve):
     return int(math.ceil(curve.key_size / 8.00))
+
+
+class AbstractFromToBytes(ABC):
```
KPrasch: Hmm - This name needs is too literal (We dont call a car a wheel!).  I'm not sure what an "AbstractFromToBytes" is.  In past team conversations about high level objects, we discussed "CryptoEntity", thoughts?

tuxxy:
I don't think we can honestly use a generic "CryptoEntity" in any useable
manner. If this pattern is used, it needs to use something like
"SerializableBytes".

umbral/utils.py
KPrasch: Over correction, IMO. How about referring to "split-key fragment object".

umbral/utils.py
```diff
@@ -42,3 +44,23 @@ def kdf(ecpoint, key_length):
...
+        Instantiate a KFrag object from the serialized data.
+        """
+        pass
+
+    @abstractmethod
+    def to_bytes(self):
+        """
+        Serialize the KFrag into a bytestring.
+        """
+        pass
```
KPrasch: Hmm, as per the official python docs, this function needs to raise `NotImplementedError`.  See: https://docs.python.org/3/library/exceptions.html#NotImplementedError

umbral/utils.py
KPrasch: Agreed - I cannot think of any reason why `from_bytes` is not a `@classmethod`.

umbral/utils.py
kikofernandez: thanks! I didn't know about this

umbral/utils.py
kikofernandez: I have called it `AbstractCryptoEntity` for now. I am not familiar with the domain and will go with whatever name seems best to the team

kikofernandez:
@KPrasch I have squashed all changes under a single commit.

jMyles:
OK, so let's look seriously at `AbstractCryptoEntity`.  What is this thing?

Is it a base class for facilitating reuse of critical logic inherent in cryptographic communication?  If so, it's odd that it is the base for multiple levels of logic - for example, `UmbralPublicKey` is a composition of `Point`, but both rely on the same base class?  That seems odd.

While it may "work" in the limited sense that both these classes need to expose de/serialization, it seems that its usefulness is likely to end there if indeed it is perceived as a base class for both of these levels.

Let me describe one use case for a `CryptoEntity` base class that seems to be precluded by this pattern:

I want `CryptoEntities` to have awareness of their constituency.  For example, `Capsule` has a constituency of five `Points` and one `BigNum`.  It is possible then to determine the bytes-serialized length of a `Capsule` (for a given curve) on this basis.  Now *that* is a useful feature in my mind!  (Although, FWIW, there is ongoing internal debate on this).

On the other hand, if these primitive constituents themselves (again, `Point` and `BigNum`) are of the same class, then we've lost the ability to reason about that topic. 

In your mind, what is a `AbstractCryptoEntity`?  Maybe this question is best answered with a docstring - I think that this class needs one.

Right now, I see a medium-term need for *some* shared logic between `Capsule`, `KFrag`, `CFrag`, and `UmbralPublicKey`.  However, cajoling `UmbralPrivateKey`, `Point`, and `BigNum` into this same foray makes me think that we're perhaps going down a wrong path here.

Whether we ultimately merge this work or not, I want to thank you for bringing this discussion to the forefront.  You might consider also engaging on Discord?

michwill:
I'd say, this class doesn't have anything to do with _crypto_. So there are two questions here:

1. Do we want to make abstract classes like this which enforce all the functionality which has to be implemented? It's not necessarily a bad way: Zope did it to some degree (way before abc), and then, I think, `zope.interface` is still used in pyramid framework. The question is: do we find this useful and convenient pattern at all?

2. If we do find it useful, it's obviously not a cryptothing, but a class which enforces specifying bytes serialization/deserialization. Is there a better name for that?

kikofernandez:
@jMyles I didn't know about Discord. In any case, my family situation (I am on parental leave) makes me really unreliable to talk because you will listen my 10-month son yelling stuff :) I wrote this code when he takes naps from the mobile phone.

@michwill I love types, type theory and design patterns. To me, if your API is stable enough I would have no doubts about using this pattern. Maybe the code is not that big now but it will help your new hires (in the future) to not remove/rename methods that should be there, among other things. IMHO, you are using a safety net :)

tuxxy:
This is something we want to work on to mature the library, but it's not something we're ready to tackle now. There needs to be more thought as to the future architecture of this library.

Issue #98, opened 2018-03-29
Generate Umbral keys with a KDF


cygnusv:
Currently, Umbral generates encrypting keypairs randomly. That's fine, but we should also support KDF-based generation, since we are going to need it for tagged capsules.

PR #111, opened 2018-04-09
Derive UmbralPrivateKeys from labels

cygnusv:
- Created a new class called `UmbralKeyingMaterial` that handles keying material for Umbral and that allows deterministic derivation of `UmbralPrivateKeys` based on labels. 
- Adds to `UmbralKeyingMaterial` a method to create `UmbralPrivateKeys` from a label (Solves #98), which will be used to support per-label Capsules. 
- Creates an abstract class `AbstractUmbralKeyingMaterial` that generalizes common behavior of `UmbralKeyingMaterial` and `UmbralPrivateKey` (i.e., secure serialization and deserialization)
- Refines the `hash_to_bn` function, which was relying on a try/except pattern for classifying the types of inputs. 
- Tests for key derivation based on labels

### DIFF
--- a/tests/test_keys/test_umbral_keys.py
+++ b/tests/test_keys/test_umbral_keys.py
@@ -95,3 +95,21 @@ def test_umbral_key_to_cryptography_keys():
     umbral_affine = umbral_pub_key.point_key.to_affine()
     x, y = crypto_pubkey.public_numbers().x, crypto_pubkey.public_numbers().y
     assert umbral_affine == (x, y)
+
+def test_keying_material_serialization():
+    umbral_keying_material = keys.UmbralKeyingMaterial()
+
+    encoded_key = umbral_keying_material.to_bytes()
+
+    decoded_key = keys.UmbralKeyingMaterial.from_bytes(encoded_key)
+    assert umbral_keying_material.keying_material == decoded_key.keying_material
+
+
+def test_keying_material_serialization_with_encryption():
+    
+    umbral_keying_material = keys.UmbralKeyingMaterial()
+
+    encoded_key = umbral_keying_material.to_bytes(password=b'test')
+
+    decoded_key = keys.UmbralKeyingMaterial.from_bytes(encoded_key, password=b'test')
+    assert umbral_keying_material.keying_material == decoded_key.keying_material
--- a/tests/test_keys/test_umbral_keys.py
+++ b/tests/test_keys/test_umbral_keys.py
@@ -9,6 +9,32 @@ def test_gen_key():
     umbral_pub_key = umbral_priv_key.get_pubkey()
     assert type(umbral_pub_key) == keys.UmbralPublicKey
 
+def test_derive_key_from_label():
+    master_secret = b"random master secret"
+    label = b"my_healthcare_information"
+
+    priv_key1 = keys.UmbralPrivateKey.derive_key_from_label(master_secret, label)
+    assert type(priv_key1) == keys.UmbralPrivateKey
+
+    pub_key1 = priv_key1.get_pubkey()
+    assert type(pub_key1) == keys.UmbralPublicKey
+
+
+    priv_key2 = keys.UmbralPrivateKey.derive_key_from_label(master_secret, label)
+    pub_key2 = priv_key2.get_pubkey()
+    assert priv_key1.bn_key == priv_key2.bn_key
+    assert pub_key1 == pub_key2
+
+    # A salt can be used too, but of course it affects the derived key
+    salt = b"optional, randomly generated salt"
+    priv_key3 = keys.UmbralPrivateKey.derive_key_from_label(master_secret, label, salt=salt)
+    assert priv_key3.bn_key != priv_key1.bn_key
+
+    # Different labels on the same master secret create different keys
+    label = b"my_tax_information"
+    priv_key4 = keys.UmbralPrivateKey.derive_key_from_label(master_secret, label)
+    pub_key4 = priv_key4.get_pubkey()
+    assert priv_key1.bn_key != priv_key4.bn_key
 
 def test_private_key_serialization(random_ec_bignum1):
     priv_key = random_ec_bignum1
--- a/tests/test_keys/test_umbral_keys.py
+++ b/tests/test_keys/test_umbral_keys.py
@@ -21,26 +21,26 @@ def test_derive_key_from_label():
 
     label = b"my_healthcare_information"
 
-    priv_key1 = umbral_keying_material.derive_private_key_from_label(label)
+    priv_key1 = umbral_keying_material.derive_privkey_by_label(label)
     assert type(priv_key1) == keys.UmbralPrivateKey
 
     pub_key1 = priv_key1.get_pubkey()
     assert type(pub_key1) == keys.UmbralPublicKey
 
     # Check that key derivation is reproducible
-    priv_key2 = umbral_keying_material.derive_private_key_from_label(label)
+    priv_key2 = umbral_keying_material.derive_privkey_by_label(label)
     pub_key2 = priv_key2.get_pubkey()
     assert priv_key1.bn_key == priv_key2.bn_key
     assert pub_key1 == pub_key2
 
     # A salt can be used too, but of course it affects the derived key
     salt = b"optional, randomly generated salt"
-    priv_key3 = umbral_keying_material.derive_private_key_from_label(label, salt=salt)
+    priv_key3 = umbral_keying_material.derive_privkey_by_label(label, salt=salt)
     assert priv_key3.bn_key != priv_key1.bn_key
 
     # Different labels on the same master secret create different keys
     label = b"my_tax_information"
-    priv_key4 = umbral_keying_material.derive_private_key_from_label(label)
+    priv_key4 = umbral_keying_material.derive_privkey_by_label(label)
     pub_key4 = priv_key4.get_pubkey()
     assert priv_key1.bn_key != priv_key4.bn_key
 
--- a/tests/test_keys/test_umbral_keys.py
+++ b/tests/test_keys/test_umbral_keys.py
@@ -10,29 +10,31 @@ def test_gen_key():
     assert type(umbral_pub_key) == keys.UmbralPublicKey
 
 def test_derive_key_from_label():
-    master_secret = b"random master secret"
+    
+    umbral_keying_material = keys.UmbralKeyingMaterial()
+
     label = b"my_healthcare_information"
 
-    priv_key1 = keys.UmbralPrivateKey.derive_key_from_label(master_secret, label)
+    priv_key1 = umbral_keying_material.derive_private_key_from_label(label)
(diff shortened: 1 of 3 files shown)


jMyles:
leaky-chimney

tests/test_keys/test_umbral_keys.py
```diff
@@ -9,6 +9,32 @@ def test_gen_key():
     umbral_pub_key = umbral_priv_key.get_pubkey()
     assert type(umbral_pub_key) == keys.UmbralPublicKey
 
+def test_derive_key_from_label():
+    master_secret = b"random master secret"
```
tuxxy: This is a fine test in that it shows a key can be derived from arbitrary data a la `master_secret`.

However, I don't think this test is showing proper or secure behavior. A user might think that they can derive a key from low entropy data such as an alphanumeric password.

I don't think this is the secure/proper way to do this, see the actual derivation function below for an alternative.

umbral/keys.py
```diff
@@ -38,6 +42,26 @@ def gen_key(cls, params: UmbralParameters=None):
         bn_key = BigNum.gen_rand(params.curve)
         return cls(bn_key, params)
 
+    @classmethod
+    def derive_key_from_label(cls, master_secret: bytes, label: bytes,
```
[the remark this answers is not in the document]
tuxxy: Another option is to move this `derive_key_from_label` method onto `UmbralPrivateKey`.

This way, a user can load up a "master key" they have stored and derive keys from labels in a more simplified API on the key.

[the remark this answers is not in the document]
cygnusv: While I completely agree that we should make this as misuse resistance as possible and the problems associated to an arbitrary bytes input, I think that allowing UmbralPrivateKeys to be used as master secret to other derived keys is relatively dangerous. The problem I see is that if we allow to use the "master" UmbralPrivateKey to be used somewhere (e.g., to create KFrags), then in the event of a leakage or a collusion attack, all the derived keys are compromised too. 

So, some alternatives:
1) Creating a MasterSecretKey class to represent a high-entropy random sequence of bytes, and use this to derive UmbralPrivateKeys

2) Creating a special subclass of UmbralPrivateKey that somehow doesn't allow to be used by Umbral directly

Thoughts?

tests/test_keys/test_umbral_keys.py
cygnusv: Agree.

umbral/keys.py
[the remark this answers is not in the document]
jMyles: @cygnusv: This pattern sounds good to me.

However, I have never liked the name "MasterKey" (or in this case, "MasterSecretKey") because, even in a world where so many things are called "keys" even though they haven't even a vaguely coherent metaphor with an unlocking device, this one is particularly stretched.  If it will never be used as a key, how about:

```
class MasterSecret(UmbralPrivateKey):
    ...
```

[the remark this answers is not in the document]
cygnusv: Oh yes, you are right about the name. I'm probably too promiscuous with the work "key".
Now that I think of, it probably has more sense that MasterSecret (or Secret or whatever) is a superclass...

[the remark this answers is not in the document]
jMyles: Yeah?  But isn't this what we want the implementer not to do?  Lest they re-use a key per @cygnusv's concern above?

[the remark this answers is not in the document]
cygnusv: OK, I see there are two approaches here:

1) There is a conversion path between `MasterSecret` and `UmbralPrivateKey` (in one direction, the other, or both). This not only includes explicit methods (e.g., `UmbralPrivateKey.to_master_secret()`) but also inheritance relations (e.g., if `MasterSecret` is subclass of `UmbralPrivateKey`, then trivially a `MasterSecret` can already be used as `UmbralPrivateKey`). The problem with this approach is that if there's a way to use the same secret material as an `UmbralPrivateKey`, then we have the problems I mentioned before in case of key leakage. I don't see a clear advantage of this approach, apart from the fact that the users only need to care about one of the classes. 

2) There is no such conversion path. If we go in this direction, then there is no way for the implementer to reuse secret material between `MasterSecrets` and `UmbralPrivateKeys`. This is probably the safest approach.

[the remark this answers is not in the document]
jMyles: I don't see any advantages of the former.  If there aren't any, let's go with the latter.

tests/test_keys/test_umbral_keys.py
michwill: What if we do something like:
```
b'very secure master seed approve recall age route small fine gospel depend canvas inherit impulse people'
```
(yes, this was a secure seed until I posted it)

umbral/bignum.py
```diff
@@ -253,11 +254,16 @@ def __hash__(self):
 
 
 def hash_to_bn(crypto_items, params):
+    if not isinstance(crypto_items, list):
+        crypto_items = [crypto_items]
```
tuxxy: We should be using argument packing here.

It would look like this:
`def hash_to_bn(params, *crypto_items)`

This packs all the `crypto_items` into a tuple that you can iterate over. You can call it like this:
`new_bn = hash_to_bn(params, x_1, x_2, x_3, x_4, ...)`

Then you don't have to worry about it being an iterable.

umbral/bignum.py
```diff
@@ -253,11 +254,16 @@ def __hash__(self):
...
+
+    valid_instance_types = (BigNum, umbral.point.Point)
+
     blake2b = hashes.Hash(hashes.BLAKE2b(64), backend=backend)
     for item in crypto_items:
-        try:
+        if isinstance(item, valid_instance_types):
```
tuxxy: The try-block is more pythonic here. Is there any reason why you are wanting to switch it?

umbral/keys.py
```diff
@@ -10,13 +10,78 @@
...
 
+from abc import ABC, abstractmethod
+ 
+class AbstractUmbralKeyingMaterial(ABC):
+
+    @abstractmethod
+    def _get_keying_material(self):
+        pass
+
+    def to_bytes(self, password: bytes=None, _scrypt_cost: int=20):
```
tuxxy: Looks like you haven't merged the latest changes where I also pass in an encoder to allow arbitrary encoding of the bytes. See #112

umbral/keys.py
```diff
@@ -10,13 +10,78 @@
...
 
 from umbral.config import default_params
-from umbral.point import Point, BigNum
+from umbral.point import Point
+from umbral.bignum import BigNum, hash_to_bn
 from umbral.params import UmbralParameters
 
+from abc import ABC, abstractmethod
+ 
+class AbstractUmbralKeyingMaterial(ABC):
```
tuxxy: I don't think an abstract base class is needed here. `UmbralKeyingMaterial` can be made as its own independent class. I think it's far more readable that way.

umbral/keys.py
```diff
@@ -239,3 +261,61 @@ def __eq__(self, other):
...
+                    salt=salt,
+                    info=b"NuCypherKMS/KeyDerivation/"+label,
+                    backend=default_backend()
+                    )
+
+        bn_key = hash_to_bn(hkdf.derive(self.keying_material), params)
+        return UmbralPrivateKey(bn_key, params)
+
+    @classmethod
+    def from_bytes(cls, key_data: bytes, password: bytes=None, _scrypt_cost: int=20):
```
tuxxy: Let's keep `to_bytes` and `from_bytes` in `UmbralPrivateKey` so that it matches with `UmbralPublicKey`.

If we want to make a move to put these in a base class, we need to think more about structuring the rest of the key classes like that. Either way, I think something like this should be in its own independent PR.

umbral/keys.py
```diff
@@ -239,3 +261,61 @@ def __eq__(self, other):
...
+    def __init__(self, keying_material: bytes=None):
+        """
+        Initializes an UmbralKeyingMaterial.
+        """
+        if keying_material:
+            if len(keying_material) < 32:
+                raise ValueError("UmbralKeyingMaterial must have size at least 32 bytes.")
+            self.keying_material = keying_material
+        else:
+            self.keying_material = os.urandom(32)
```
tuxxy: I believe that Blake2b can fit 64 bytes of entropy in it. Perhaps we generate 64 bytes from urandom and utilize all the available entropy.

umbral/keys.py
```diff
@@ -10,13 +10,78 @@
...
+            ).derive(password)
+
+            umbral_get_keying_material = SecretBox(key).encrypt(umbral_get_keying_material)
+            umbral_get_keying_material += salt
+
+        encoded_key = base64.urlsafe_b64encode(umbral_get_keying_material)
+        return encoded_key
+
+    @classmethod
+    def _decode_get_keying_material(cls, key_data: bytes,
```
tuxxy: This method name is a little funky. I think I prefer simply `_get_keying_material`.

umbral/keys.py
```diff
@@ -239,3 +261,61 @@ def __eq__(self, other):
...
+            if len(keying_material) < 32:
+                raise ValueError("UmbralKeyingMaterial must have size at least 32 bytes.")
+            self.keying_material = keying_material
+        else:
+            self.keying_material = os.urandom(32)
+
+    def _get_keying_material(self):
+        return self.keying_material
+
+    def derive_private_key_from_label(self, label: bytes,
```
tuxxy: I think I prefer `derive_privkey_by_label`. We have taken the standard to calling all keys `privkey` or `pubkey` in one word without the underscore.

umbral/bignum.py
tuxxy: Perhaps this can also be in a different PR, too?

umbral/bignum.py
cygnusv: OK, I'll revert this

umbral/keys.py
cygnusv: OK, great. I'm all in for reducing verbosity.

umbral/bignum.py
cygnusv: OK, I will revert this changes. Argument packing makes sense, I will do that in a different PR

umbral/keys.py
cygnusv: OK

umbral/keys.py
tuxxy: Sounds good.

tests/test_keys/test_umbral_keys.py
```diff
@@ -15,6 +15,34 @@ def test_gen_key():
     umbral_pub_key = umbral_priv_key.get_pubkey()
     assert type(umbral_pub_key) == keys.UmbralPublicKey
 
+def test_derive_key_from_label():
+    
+    umbral_keying_material = keys.UmbralKeyingMaterial()
+
+    label = b"my_healthcare_information"
+
+    priv_key1 = umbral_keying_material.derive_privkey_by_label(label)
```
michwill: `derive_privkey_by_label` looks a little lengthy, but I cannot think of a better name for the method easily

PR #112, opened 2018-04-10
Add encoder/decoder params to to_bytes and from_bytes for Umbral keys

tuxxy:
### What this does:
1. Adds `encoder` and `decoder` params to `to_bytes` and `from_bytes` on `UmbralPrivateKey` and `UmbralPublicKey` for more explicit serialization.
2. Adds a test `test_key_encoder_decoder` to `test_umbral_keys`.

This allows for more explicit and flexible serialization when it comes to Umbral keys.

### DIFF
--- a/umbral/keys.py
+++ b/umbral/keys.py
@@ -1,7 +1,7 @@
 import os
 import base64
 
-from typing import Union, Callable
+from typing import Callable
 
 from nacl.secret import SecretBox
 from cryptography.hazmat.primitives.asymmetric import ec
@@ -43,7 +43,9 @@ class UmbralPrivateKey(object):
                    password: bytes=None, _scrypt_cost: int=20,
                    decoder: Callable=None):
         """
-        Loads an Umbral private key from a urlsafe base64 encoded string.
+        Loads an Umbral private key from bytes.
+        Optionally, allows a decoder function to be passed as a param to decode
+        the data provided before converting to an Umbral key.
         Optionally, if a password is provided it will decrypt the key using
         nacl's Salsa20-Poly1305 and Scrypt key derivation.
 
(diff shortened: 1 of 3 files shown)


umbral/keys.py
```diff
@@ -39,8 +39,9 @@ def gen_key(cls, params: UmbralParameters=None):
         return cls(bn_key, params)
 
     @classmethod
-    def from_bytes(cls, key_data: bytes, params: UmbralParameters=None,
-                 password: bytes=None, _scrypt_cost: int=20):
+    def from_bytes(cls, key_bytes: bytes, params: UmbralParameters=None,
+                   password: bytes=None, _scrypt_cost: int=20,
+                   decoder: Callable=None):
         """
         Loads an Umbral private key from a urlsafe base64 encoded string.
```
KPrasch: Stale comment here wrt the encoder function.

umbral/keys.py
```diff
@@ -54,7 +55,8 @@ def from_bytes(cls, key_data: bytes, params: UmbralParameters=None,
         if params is None:
             params = default_params()
 
-        key_bytes = base64.urlsafe_b64decode(key_data)
+        if decoder:
```
KPrasch: So - if the decoder is `None` we don't use one at all?  Do we want to provide a default?

umbral/keys.py
```diff
@@ -99,11 +102,13 @@ def to_bytes(self, password: bytes=None, _scrypt_cost: int=20):
                 backend=default_backend()
             ).derive(password)
 
-            umbral_priv_key = SecretBox(key).encrypt(umbral_priv_key)
-            umbral_priv_key += salt
+            umbral_privkey = SecretBox(key).encrypt(umbral_privkey)
+            umbral_privkey += salt
```
KPrasch: :+1:

umbral/keys.py
```diff
@@ -166,29 +173,30 @@ def __init__(self, point_key, params: UmbralParameters=None):
...
+        if decoder:
+            key_bytes = decoder(key_bytes)
 
         point_key = Point.from_bytes(key_bytes, params.curve)
         return cls(point_key, params)
 
-    def to_bytes(self):
+    def to_bytes(self, encoder: Callable=None):
         """
         Returns an Umbral public key as a urlsafe base64 encoded string.
```
KPrasch: stale?

PR #113, opened 2018-04-11
Correcting staticmethod to classmethod

kikofernandez:
It seems like the `from_bytes` method in `KFrag` and `CapsuleFrag` classes should have been a `@classmethod` but it currently is a `@staticmethod`. I removed unused imports as well.


kikofernandez:
Thanks for accepting the PR! 
Awesome!

Issue #100, opened 2018-03-31
Integrate the challenge protocol as part of the re-encryption request/reply 


cygnusv:
Instead of having a separate protocol for verifying correctness ("the challenge protocol") we can simply integrate it as part of the re-encryption request and reply.

Basically, the idea is to include the challenge request as part of/together with the `Capsule`, and put the challenge response as part of the `cfrag`. The disadvantage of that is that we are adding a communications overhead for all re-encryption requests (i.e., metadata such as timestamp to avoid replays) and re-encryption responses (i.e., around 5 or 6 points/bignums). The advantage is that we get rid of supporting an additional protocol, and on top of that, we always get correctness guarantees on each re-encryption.

tuxxy:
This was resolved in #115, so I'm closing it. Thanks, @cygnusv!

PR #115, opened 2018-04-17
Re-encryption with correctness proofs

### LINKED ISSUE #120: Define format for re-encryption metadata?
As part of the re-encryption correctness proof (which in PR#111 is always included together the result of the re-encryption), it is possible to take some optional "metadata" information into consideration. This "metadata" is intended to serve as additional information that binds the re-encryption request with its external context. 

For example, in the case of NuCypher KMS, this metadata can include a timestamp (so that Ursula cannot respond with an old re-encryption and/or correctness proof), the policy ID (so Ursula cannot reuse work done for other policies, even for the same capsule, Alice & Bob), etc. 

Do we want to leave this format arbitrary? This has implications wrt to (de)serialization of `CorrectnessProofs` and, consequently, of `CFrags`

cygnusv:
# What this does:
- Renames `ChallengeResponse` to `CorrectnessProof`, and integrates it inside `CFrags` (Solves #100)
- Re-encryption now produces correctness proofs by default, although this can be disabled with the parameter `provide_proof`.
- Re-encryption correctness is verified during decryption. If it fails, a new type of exception called `UmbralCorrectnessError` is raised, which includes a list of the `CFrags` that don't pass the correctness verification (so that Bob can take the action he considers more convenient).
- Metadata is now integrated into the `CorrectnessProof` (and hence, into the `CFrag`). As a consequence, metadata is now a parameter of `pre.reencrypt()` (although optional). 
- Metadata can be any arbitrary `bytes` object (Closes #120)
- Tests for everything


tests/test_keys/test_key_fragments.py
```diff
@@ -19,20 +20,103 @@ def test_kfrag_serialization(alices_keys):
...
     _unused_key, capsule = pre._encapsulate(pub_key_alice.point_key)
-    k_frags = pre.split_rekey(priv_key_alice, pub_key_alice, 1, 2)
+    kfrags = pre.split_rekey(priv_key_alice, pub_key_alice, 1, 2)
+
+    # Example of potential metadata to describe the re-encryption request
+    metadata = { 'ursula_id' : 42, 
+                 'timestamp' : time.time(), 
+                 'capsule' : bytes(capsule), 
+               }
+    metadata = str(metadata).encode()
```
tuxxy: Obviously, we want a better form of serialization than just turning a dict into a string.

I presume this is being blocked due to not having a format for metadata in the KMS?

umbral/fragments.py
```diff
@@ -75,12 +75,79 @@ def __bytes__(self):
...
+        e2 = Point.from_bytes(data.read(key_size + 1), curve)
+        v2 = Point.from_bytes(data.read(key_size + 1), curve)
+        kfrag_commitment = Point.from_bytes(data.read(key_size + 1), curve)
+        kfrag_pok = Point.from_bytes(data.read(key_size + 1), curve)
+        kfrag_sig1 = BigNum.from_bytes(data.read(key_size), curve)
+        kfrag_sig2 = BigNum.from_bytes(data.read(key_size), curve)
+        sig = BigNum.from_bytes(data.read(key_size), curve)
+
+        metadata = data.read()
+        if metadata == bytes(0):
```
tuxxy: This is build a bit weirdly.

You can simply do `metadata = data.read() or None`. If there is no data, it will be `b''` which loses precedence to `None`, then you don't have to set it to `None`.

That is unless you are setting `metadata` to be 0x00 and it will always have that byte, but I don't like that.

umbral/fragments.py
```diff
@@ -75,12 +75,79 @@ def __bytes__(self):
...
+        result = e2            \
+            + v2               \
+            + kfrag_commitment \
+            + kfrag_pok        \
+            + kfrag_sig1       \
+            + kfrag_sig2       \
+            + sig              
+
+        if self.metadata is not None:
+            result = result + self.metadata
```
tuxxy: You can also do `result += self.metadata or b''`. This will concat the empty bytestring if `self.metadata` is `None` and removes the need for L135.

umbral/fragments.py
```diff
@@ -98,7 +165,10 @@ def from_bytes(cls, data: bytes, curve: ec.EllipticCurve = None):
         kfrag_id = BigNum.from_bytes(data.read(key_size), curve)
         eph_ni = Point.from_bytes(data.read(key_size + 1), curve)
 
-        return cls(e1, v1, kfrag_id, eph_ni)
+        proof = data.read()
+        proof = CorrectnessProof.from_bytes(proof, curve) if proof != bytes(0) else None
```
tuxxy: Same logic applies here from what I listed above.

umbral/fragments.py
```diff
@@ -109,7 +179,14 @@ def to_bytes(self):
         kfrag_id = self.bn_kfrag_id.to_bytes()
         eph_ni = self.point_eph_ni.to_bytes()
 
-        return e1 + v1 + kfrag_id + eph_ni
+        serialized_cfrag = e1 + v1 + kfrag_id + eph_ni
+
+        if self.proof is not None:
+            serialized_cfrag += self.proof.to_bytes()
```
tuxxy: As well as here.

umbral/pre.py
```diff
@@ -381,34 +342,36 @@ def _check_challenge(capsule: Capsule, cfrag: CapsuleFrag,
...
-
-    return check31 & check32 & check33 & check34
-
+    valid_kfrag_signature     = z1 == BigNum.hash_to_bn(*[g_y, kfrag_id, pub_a, pub_b, u1, xcomp], params=params)
+    correct_reencryption_of_e = z3 * e == e2 + (h * e1)
+    correct_reencryption_of_v = z3 * v == v2 + (h * v1)
+    correct_rk_commitment     = z3 * u == u2 + (h * u1)
+    
+    return valid_kfrag_signature        \
+         & correct_reencryption_of_e    \
```
tuxxy: These are bitwise ANDs in Python. They accomplish the same goal, but I think it's more Pythonic to use `and` instead.

umbral/pre.py
```diff
@@ -381,34 +342,36 @@ def _check_challenge(capsule: Capsule, cfrag: CapsuleFrag,
...
     h = BigNum.hash_to_bn(*hash_input, params=params)
 
-    check31 = z1 == BigNum.hash_to_bn(g_y, kfrag_id, pub_a, pub_b, u1, xcomp, params=params)
-    check32 = z3 * e == e2 + (h * e1)
-    check33 = z3 * u == u2 + (h * u1)
-    check34 = z3 * v == v2 + (h * v1)
-
-    return check31 & check32 & check33 & check34
-
+    valid_kfrag_signature     = z1 == BigNum.hash_to_bn(*[g_y, kfrag_id, pub_a, pub_b, u1, xcomp], params=params)
```
tuxxy: You don't need to do `*[g_y, kfrag_id, pub_a, pub_b, u1, xcomp]` here.
You can simply do: `BigNum.hash_to_bn(g_y, kfrag_id, pub_a, pub_b, u1, xcomp, params=params)`.

tuxxy: Furthermore, not sure if I like doing this oracle operation and the check on the same line. I'll defer to other opinions on this, though.

umbral/pre.py
```diff
@@ -497,8 +460,8 @@ def encrypt(alice_pubkey: UmbralPublicKey, plaintext: bytes,
     return ciphertext, capsule
 
 
-def _open_capsule(capsule: Capsule, bob_private_key: UmbralPrivateKey,
-                  alice_pub_key: UmbralPublicKey, params: UmbralParameters=None) -> bytes:
+def _open_capsule(capsule: Capsule, bob_privkey: UmbralPrivateKey,
```
tuxxy: 👍

umbral/pre.py
```diff
@@ -507,10 +470,19 @@ def _open_capsule(capsule: Capsule, bob_private_key: UmbralPrivateKey,
     """
     params = params if params is not None else default_params()
 
-    priv_b = bob_private_key.bn_key
-    pub_b = priv_b * params.g
+    priv_b = bob_privkey.bn_key
+    pub_b = bob_privkey.get_pubkey().point_key
```
tuxxy: If my PR gets in before this one, you can do `bob_privkey.pubkey.point_key` because it gets cached.

umbral/pre.py
```diff
@@ -507,10 +470,19 @@ def _open_capsule(capsule: Capsule, bob_private_key: UmbralPrivateKey,
...
 
-    pub_a = alice_pub_key.point_key
+    pub_a = alice_pubkey.point_key
+
+    # TODO: Change dict for a list if issue #116 goes through
```
tuxxy: Is this an old TODO?

umbral/pre.py
```diff
@@ -326,12 +277,16 @@ def reencrypt(kfrag: KFrag, capsule: Capsule,
...
+    if provide_proof:
+        _prove_correctness(cfrag, kfrag, capsule, metadata, params)
+
     return cfrag
 
 
-def _challenge(kfrag: KFrag, capsule: Capsule, 
-              cfrag: CapsuleFrag, challenge_metadata: bytes=None,
-              params: UmbralParameters=None) -> ChallengeResponse:
+def _prove_correctness(cfrag: CapsuleFrag, kfrag: KFrag, capsule: Capsule,
```
tuxxy: I think `_prove_correctness` has to be above `reencrypt` for it to be used?

tuxxy: Might it also be better as a classmethod on `CorrectnessProof`?

umbral/pre.py
```diff
@@ -24,6 +24,10 @@
 class GenericUmbralError(Exception):
     pass
 
+class UmbralCorrectnessError(GenericUmbralError):
+    def __init__(self, message, offending_cfrags):
+        super().__init__(message)
+        self.offending_cfrags = offending_cfrags
```
tuxxy: Is there a need to capture these here even if we handle them ~L485?

tests/test_keys/test_key_fragments.py
cygnusv: Well, to me this is just a rough example for testing purposes. We want something reasonable, beyond a simple `b"this is the metadata of a request"`, but without being something too complex.

umbral/fragments.py
cygnusv: Fixed

umbral/fragments.py
cygnusv: Fixed

umbral/fragments.py
cygnusv: Fixed

umbral/pre.py
cygnusv: My intention was to avoid short-circuiting to make that constant-time

umbral/pre.py
cygnusv: Fixed

umbral/fragments.py
cygnusv: mmm, no, it fails when `self.proof` is None. I'll revert it to the previous way.

umbral/fragments.py
cygnusv: Fixed

umbral/pre.py
cygnusv: No, I added it here in light of issue #116

umbral/pre.py
cygnusv: The idea is to include them in the exception, so in the application (i.e., the KMS), Bob could know which cfrags were incorrect and act consequently.

tests/test_correctness.py
```diff
@@ -0,0 +1,216 @@
...
+                     'timestamp' : time.time(), 
+                     'capsule' : bytes(capsule), 
+                   }
+
+        metadata = str(metadata).encode()
+
+        cfrag = pre.reencrypt(kfrag, capsule, metadata=metadata)
+        capsule.attach_cfrag(cfrag)
+
+        assert pre._verify_correctness(capsule, cfrag,
```
michwill: I like the simplicity of the interface!

tests/test_correctness.py
```diff
@@ -0,0 +1,216 @@
...
+    for cfrag_i, metadata_i in zip(cfrags[1:], metadata[1:]):
+        assert pre._verify_correctness(capsule_alice1,
+                                       cfrag_i,
+                                       pub_key_alice.point_key,
+                                       pub_key_bob.point_key,
+                                      )
+
+    # Alternatively, we can try to open the capsule directly.
+    # We should get an exception with an attached list of incorrect cfrags
+    with pytest.raises(pre.UmbralCorrectnessError) as exception_info:
```
jMyles: Yeah, groovy.

umbral/pre.py
```diff
@@ -349,28 +304,34 @@ def _challenge(kfrag: KFrag, capsule: Capsule,
...
-                    challenge_resp: ChallengeResponse, 
-                    pub_a: Point, pub_b: Point, challenge_metadata: bytes=None,
+def _verify_correctness(capsule: Capsule, cfrag: CapsuleFrag,
+                    pub_a: Point, pub_b: Point, 
                     params: UmbralParameters=None) -> bool:
+    
+    try:
+        proof = cfrag.proof
+    except AttributeError:
+        raise ValueError("CFrag doesn't have a correctnes proof attached")
```
jMyles: Untested - needs a test I think.

umbral/fragments.py
```diff
@@ -132,8 +130,7 @@ def to_bytes(self) -> bytes:
             + kfrag_sig2       \
             + sig              
 
-        if self.metadata is not None:
-            result = result + self.metadata
+        result += self.metadata or b''
```
jMyles: This is an odd line - the user is not allowed to use something falsey as metadata?

tests/test_keys/test_key_fragments.py
cygnusv: OK, let's go the the initial approach then.

tests/test_correctness.py
```diff
@@ -0,0 +1,216 @@
+import pytest
+
+from umbral import pre, keys
+from umbral.point import Point
+from umbral.fragments import CorrectnessProof
```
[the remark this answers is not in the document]
cygnusv: I was using it through `pre.CorrectnessProof`, which was it's previous location. I forgot to change it. Now it's fixed.

PR #118, opened 2018-04-18
Implement arg packing in hash_to_bn

tuxxy:
### What this does:
1. Implements arg packing in `BigNum.hash_to_bn`
2. Moves `hash_to_bn` to be a class method on `BigNum`.
3. Raise `TypeError` in `hash_to_bn` when something can't be serialized to bytes, or isn't bytes already.

### DIFF
--- a/umbral/bignum.py
+++ b/umbral/bignum.py
@@ -6,6 +6,7 @@ from cryptography.hazmat.primitives import hashes
 
 from umbral.config import default_curve
 from umbral.utils import get_curve_keysize_bytes
+import umbral.point
 
 
 class BigNum(object):
@@ -253,11 +254,16 @@ class BigNum(object):
 
 
 def hash_to_bn(crypto_items, params):
+    if not isinstance(crypto_items, list):
+        crypto_items = [crypto_items]
+
+    valid_instance_types = (BigNum, umbral.point.Point)
+
     blake2b = hashes.Hash(hashes.BLAKE2b(64), backend=backend)
     for item in crypto_items:
-        try:
+        if isinstance(item, valid_instance_types):
             data_bytes = item.to_bytes()
-        except AttributeError:
+        else:
             data_bytes = item
         blake2b.update(data_bytes)
 
--- a/umbral/bignum.py
+++ b/umbral/bignum.py
@@ -6,7 +6,6 @@ from cryptography.hazmat.primitives import hashes
 
 from umbral.config import default_curve
 from umbral.utils import get_curve_keysize_bytes
-import umbral.point
 
 
 class BigNum(object):
@@ -252,18 +251,12 @@ class BigNum(object):
     def __hash__(self):
         return hash(int(self))
 
-
 def hash_to_bn(crypto_items, params):
-    if not isinstance(crypto_items, list):
-        crypto_items = [crypto_items]
-
-    valid_instance_types = (BigNum, umbral.point.Point)
-
     blake2b = hashes.Hash(hashes.BLAKE2b(64), backend=backend)
     for item in crypto_items:
-        if isinstance(item, valid_instance_types):
+        try:
             data_bytes = item.to_bytes()
-        else:
+        except AttributeError:
             data_bytes = item
         blake2b.update(data_bytes)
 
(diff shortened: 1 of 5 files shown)


cygnusv:
Why moving back again `hash_to_bn` to be a class method on BigNum?

My original code of `hash_to_bn` was like that, but from a discussion with @jMyles and you it was decided to leave it as a function in BigNum? Hence, my question: what made you change your mind?

tuxxy:
To be honest, I don't recall the exact nature of that discussion.

Looking at the code now, it seems like it fits better as a classmethod. It only ever returns a `BigNum`. Do you remember why we didn't like it in `BigNum`?

The only pushback I've had on this today is the name of it. :P

umbral/bignum.py
```diff
@@ -81,6 +81,34 @@ def from_int(cls, num, curve: ec.EllipticCurve=None):
...
 
+    @classmethod
+    def hash_to_bn(cls, *crypto_items, params=None):
+        params = params if params is not None else default_params()
+
+        # TODO: Clean this in an upcoming cleanup of pyUmbral
+        blake2b = hashes.Hash(hashes.BLAKE2b(64), backend=backend)
+        for item in crypto_items:
+            try:
+                item_bytes = item.to_bytes()
```
jMyles: During our discussion, we talked about only allowing bytes and nothing else.  I thought you were on board for that (although I can see the allure of taking arbitrary `crypto_items`).  Where are you standing at this point?

umbral/bignum.py
```diff
@@ -81,6 +81,34 @@ def from_int(cls, num, curve: ec.EllipticCurve=None):
 
         return cls(bignum, curve_nid, group, order)
 
+    @classmethod
+    def hash_to_bn(cls, *crypto_items, params=None):
```
jMyles: As @KPrasch pointed out, `BigNum.hash_to_bn` is very awkward.  It's not like you'd ever have `BigNum.hash_to_kfrag`.  

What is *really* happening here that makes it belong on the `BigNum` class.  It's that we're creating a `BigNum` *from* something, right?

Maybe `BigNum.one_way_from_bytes`?  Or `BigNum.hash_from_crypto_items`?  Or something to that effect?

umbral/bignum.py
```diff
@@ -81,6 +81,34 @@ def from_int(cls, num, curve: ec.EllipticCurve=None):
...
+        params = params if params is not None else default_params()
+
+        # TODO: Clean this in an upcoming cleanup of pyUmbral
+        blake2b = hashes.Hash(hashes.BLAKE2b(64), backend=backend)
+        for item in crypto_items:
+            try:
+                item_bytes = item.to_bytes()
+            except AttributeError:
+                if not isinstance(item, bytes):
```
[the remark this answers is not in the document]
jMyles: :+1:

umbral/fragments.py
```diff
@@ -66,7 +66,7 @@ def verify(self, pub_a, pub_b, params: "UmbralParameters"=None):
 
         # We check the Schnorr signature over the kfrag components
         g_y = (z2 * params.g) + (z1 * pub_a)
-        check_kfrag_2 = z1 == hash_to_bn([g_y, self.bn_id, pub_a, pub_b, u1, x], params)
+        check_kfrag_2 = z1 == BigNum.hash_to_bn(g_y, self.bn_id, pub_a, pub_b, u1, x, params=params)
```
jMyles: Actually, as much as I liked the unpacking approach, I now find myself a little anxious about the way this reads.  Maybe we want to go back to taking a collection?  I think it might be better to be really explicit here, like:

`BigNum.hash_from_crypto_items(items_to_hash=[g_y, self.bn_id, pub_a, pub_b, u1, x], params=params)`

tuxxy: Why do you find this anxiety inducing? I think it reads quite well. I don't think passing in a collection adds any sort of readability to it. They would all be inline collections, anyway. Just skip the step and pack it.

umbral/bignum.py
tuxxy: See my comment on the `bytes` thing. My response is similar. I like the idea of renaming it, but there are more important things to do before we consider this.

PR #125, opened 2018-04-23
Renaming internal attributes

cygnusv:
Renaming of attributes according to the following guidelines:

- Internal variables should not be considered public, so they should start with an underscore (e.g., `_bn_key`)
- When possible, add a prefix to the variable name that helps to understand its type (e.g., `_bn_key` is of type `BigNum`)
- When possible, use a self-explanatory name (e.g., `_bn_key`). However, although self-explanatory names are nice, avoid make them too long (e.g., `_point_noninteractive` is better than `_point_eph_noninteractive_transformation`). 
- We allow arbitrary names when there's no immediate functionality associated to the variable (e.g., `_point_e`). In this case, the preference is to tend to one-letter names, as it's customary in scientific literature.

### DIFF
--- a/tests/test_capsule/test_capsule_operations.py
+++ b/tests/test_capsule/test_capsule_operations.py
@@ -11,8 +11,8 @@ def test_capsule_creation(alices_keys):
         rare_capsule = Capsule()    # Alice cannot make a capsule this way.
 
     # Some users may create capsules their own way.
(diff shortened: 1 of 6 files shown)


jMyles:
Good stuff, but let's do #129 instead (same content as this, but without the strange merge commits)

PR #130, opened 2018-04-27
Add constant time operations 

### LINKED ISSUE #117: Some BigNum OpenSSL functions are not in constant time
### The Issue:
1. Some BigNum OpenSSL functions are _apparently_ not in constant time.
2. Cryptography.io doesn't expose a way for us to have those performed in constant time.

### What is affected:
1. `BigNum.__inverse__`
2. `BigNum.__truediv__` -- _Calls `BN_mod_inverse`_
3. `BigNum.__pow__` -- _Possibly? I'm not clear on this one. It might be constant time for all operations where the base and exponent are the size of the order of the modulus._ 

### How to fix:
1. `BN_FLG_CONSTTIME` needs to be set on all OpenSSL BIGNUMs used in Umbral.
2. `BN_set_flags` (and `BN_get_flags`) needs to be exposed to set the flag to `BN_FLG_CONSTTIME` (0x04).
3. `BN_mod_exp_mont`, `BN_mod_exp_mont_consttime`, `BN_div_no_branch`, and `BN_mod_inverse_no_branch` needs to be exposed on Cryptography.io.

### Notes:
1. It appears that pyUmbral is not affected by these timing issues due to public/non-sensitive values being used during these operations.
2. Referenced in https://github.com/pyca/cryptography/issues/4199

tuxxy:
### What this does:
1. Sets the `BN_FLG_CONSTTIME` flag by default when creating a new BN.
2. Changes `CurveBN.__pow__` to use `BN_mod_exp_mont`. When used with a `BN_FLG_CONSTTIME` BN, it calls `BN_mod_exp_mont_consttime`. Similarly, when `BN_div`, and `BN_mod_inverse` are called, it will call their constant time relatives.
3. Updates docstrings for operations that may not be in constant time.
4. Adds a context manager for montgomery operations.
5. Resolves #117


umbral/openssl.py
```diff
@@ -78,6 +80,9 @@ def _int_to_bn(py_int: int, curve_nid: int=None):
         on_curve = _bn_is_on_curve(conv_bn, curve_nid)
         if not on_curve:
             raise ValueError("The Python integer given is not on the provided curve.")
+
+    if is_consttime:
+        backend._lib.BN_set_flags(conv_bn, backend._lib.BN_FLG_CONSTTIME)
```
jMyles: Awesome, @tuxxy.  Great job shepherding this through.

umbral/openssl.py
```diff
@@ -138,3 +143,23 @@ def _get_affine_coords_via_EC_POINT(ec_point, ec_group=None, curve_nid: int=None
...
+    backend.openssl_assert(bn_mont_ctx != backend._ffi.NULL)
+    # Don't set the garbage collector. Only free it when the context is done
+    # or else you'll get a null pointer error.
+
+    with backend._tmp_bn_ctx() as bn_ctx:
+        res = backend._lib.BN_MONT_CTX_set(bn_mont_ctx, modulus, bn_ctx)
+        backend.openssl_assert(res == 1)
+        try:
+            yield bn_mont_ctx
+        finally:
```
jMyles: Love this pattern - glad to see you adopting it!  :-)

I wonder, though - do we perhaps want the try-finally *outside* the with?  Currently, if there's an unhandled exception on any of the lines above (say, due to an incorrect OpenSSL or cryptography.io version), then the `bn_mont_ctx` will never be freed.

umbral/curvebn.py
```diff
@@ -139,19 +147,19 @@ def __pow__(self, other):
...
             other = CurveBN(other, None, None, None)
 
         power = openssl._get_new_BN()
         with backend._tmp_bn_ctx() as bn_ctx:
-            res = backend._lib.BN_mod_exp(
-                power, self.bignum, other.bignum, self.order, bn_ctx
-            )
-            backend.openssl_assert(res == 1)
-
+            with openssl._tmp_bn_mont_ctx(self.order) as bn_mont_ctx:
```
jMyles: I don't love the nested context managers here - we never use the outer indentation level as a control structure.  Does it maybe make sense to either make a single context manager to handle this situation or at least throw them on the same line?

umbral/curvebn.py
```diff
@@ -13,6 +13,8 @@ class CurveBN(object):
     """
     Represents an OpenSSL Bignum modulo the order of a curve. Some of these
     operations will only work with prime numbers
+    By default, the underlying OpenSSL BIGNUM has BN_FLG_CONSTTIME set for
```
jMyles: I'm not sure this new section of docstring adds anything particularly useful, especially repeated four times.

umbral/curvebn.py
```diff
@@ -174,7 +182,7 @@ def __truediv__(self, other):
         """
         Performs a BN_div on two BIGNUMs (modulo the order of the curve).
 
-        WARNING: Not in constant time yet.
+        WARNING: Only in constant time if BN_FLG_CONSTTIME is set on the BN.
```
jMyles: Yeah, good looking out.

umbral/openssl.py
```diff
@@ -1,16 +1,19 @@
+from contextlib import contextmanager
 from cryptography.hazmat.backends.openssl import backend
 
 
-def _get_new_BN():
+def _get_new_BN(is_consttime=True):
```
jMyles: `is_consttime`?  How about `use_constant_time_operations`?

umbral/openssl.py
```diff
@@ -63,13 +66,12 @@ def _bn_is_on_curve(check_bn, curve_nid: int):
...
 
-def _int_to_bn(py_int: int, curve_nid: int=None):
+def _int_to_bn(py_int: int, curve_nid: int=None, is_consttime=True):
     """
     Converts the given Python int to an OpenSSL BIGNUM. If a curve_nid is
     provided, it will check if the Python integer is within the order of that
     curve. If it's not within the order, it will raise a ValueError.
-
-    TODO: Add flag for consttime
+    is_consttime set to True by default to use constant time ops.
```
jMyles: This part of the docstring doesn't quite make sense.  Probably better is:

`:param is_consttime (or better name): Whether to tell the backend to use constant-time operations for this instance.

umbral/openssl.py
```diff
@@ -63,13 +66,12 @@ def _bn_is_on_curve(check_bn, curve_nid: int):
     return check_sign == 1 and range_check == -1
 
 
-def _int_to_bn(py_int: int, curve_nid: int=None):
+def _int_to_bn(py_int: int, curve_nid: int=None, is_consttime=True):
```
KPrasch: Nice work, how about the name `constant_time`? :1st_place_medal:

umbral/openssl.py
tuxxy: Hmm, not sure..

If you notice that I set the BN_MONT_CTX with a modulus and also BN_CTX. My concern is that if we let the BN_CTX fall out of the context, the BN_MONT_CTX structure will fail.

I don't know how this works due to the mix between Python and C and how it would be handled, but I have a feeling that it must be used within the BN_CTX context.

umbral/openssl.py
tuxxy: Do you feel good about simply shortening it to `constant_time=True`?

`use_constant_time` kinda feels like it affects _everything_ as a BIGNUM, when it only affect a few operations.

umbral/openssl.py
jMyles: Hmm - I think you might be misunderstanding what I was saying.  I'm thinking this:

```
    try:
        with backend._tmp_bn_ctx() as bn_ctx:
            res = backend._lib.BN_MONT_CTX_set(bn_mont_ctx, modulus, bn_ctx)
            backend.openssl_assert(res == 1)
            yield bn_mont_ctx
    finally:
        backend._lib.BN_MONT_CTX_free(bn_mont_ctx)
```

...or are you saying that you have to be in the `bn_ctx` to *free* (but not init) the `bn_mont_ctx`?  I don't think that's so, although I haven't tried it.

umbral/openssl.py
jMyles: Well, conversely, `constant_time=True` feels like it only effects this instantiation, rather than the resulting object.

umbral/openssl.py
tuxxy: Oh, I see what you're saying. Yeah, you could do that.

umbral/curvebn.py
michwill: How would they be thrown into the same line?

jMyles: You can just stick a comma between them.

But I also think it probably makes sense to make a single outer context manager for this situation, no?

michwill: Hm, perhaps. Do we ever use `_tmp_bn_mont_ctx` without the other indeed?

tuxxy:
No, both are usually used together

PR #131, opened 2018-04-30
Constant time `hash_to_bn` (and some minor cleaning)

### LINKED ISSUE #110: Investigate constant-time alternatives to hash_to_bn

cygnusv:
- Implements `CurveBN.hash_to_bn()` in constant-time (resolves #110). There is a negligible probability that it requires a second call, which occurs when the digest returns 0. Now, the `params.CURVE_MINVAL_HASH` constant is not needed anymore.
- Renames `CurveBN.hash_to_bn()` as  `CurveBN.hash()`. This is a temporary decision until we find a better place to put this functionality.
- Uses Blake2b instead of SHA256 in `unsafe_hash_to_point`
- Removes `Point.get_order_from_curve`. This function was returning an ill-formed `CurveBN`, since it wasn't being used as a BigNum modulo the order of the curve. 
- Removes some unused imports

### NOTE:
- Actual constant-time requires to set the flag `BN_FLG_CONSTTIME`, as supported by PR #130 (pending at the time of this writing).

### DIFF
--- a/umbral/curvebn.py
+++ b/umbral/curvebn.py
@@ -72,7 +72,7 @@ class CurveBN(object):
         return cls(conv_bn, curve_nid, group, order)
 
     @classmethod
-    def hash_to_bn(cls, *crypto_items, params=None):
+    def hash_to_bn(cls, *crypto_items, params=None, _iteration=0):
         params = params if params is not None else default_params()
 
         # TODO: Clean this in an upcoming cleanup of pyUmbral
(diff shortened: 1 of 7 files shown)


umbral/curvebn.py
```diff
@@ -85,19 +85,34 @@ def hash_to_bn(cls, *crypto_items, params=None):
...
+                    bn_from_hash, hash_digest, order, bn_ctx
+                )
+                backend.openssl_assert(res == 1)
+
+            group = openssl._get_ec_group_by_curve_nid(curve_nid)
+            bn_from_hash = cls(bn_from_hash, curve_nid, group, order)
+
+        except ValueError:
+            # This case is only reached when the result is 0, 
+            # which happens with prob. 1/order of the curve
```
michwill: `1/order` (rather than `1 / 2 ** order`) probability could be non-negligible. Whether it is or not - I'll leave up to @cygnusv and @tuxxy to decide

cygnusv: The order of the curve is already a number in the order of `2**n`. For example, for `secp256k1`, the order of the curve is approximately `2**255.99999`

michwill: Ah. All right, in this case `1 / order` is totally fine, of course

umbral/curvebn.py
```diff
@@ -85,19 +85,34 @@ def hash_to_bn(cls, *crypto_items, params=None):
...
-        hash_bn = h % int(params.order)
-        res = cls.from_int(hash_bn, params.curve)
-        return res
+        hash_digest = blake2b.finalize()
+        hash_digest = int.from_bytes(hash_digest, byteorder='big', signed=False)
+
+        curve_nid = backend._elliptic_curve_to_nid(params.curve)
+        order = openssl._get_ec_order_by_curve_nid(curve_nid)
+
+        try:
```
KPrasch: Is this `try` block too widely scoped - which line actually raises the `ValueError`?  Perhaps `try/except/else` is useful here.

umbral/curvebn.py
```diff
@@ -85,19 +85,34 @@ def hash_to_bn(cls, *crypto_items, params=None):
...
+                backend.openssl_assert(res == 1)
+
+            group = openssl._get_ec_group_by_curve_nid(curve_nid)
+            bn_from_hash = cls(bn_from_hash, curve_nid, group, order)
+
+        except ValueError:
+            # This case is only reached when the result is 0, 
+            # which happens with prob. 1/order of the curve
+            _iteration += 1
+            bn_from_hash = cls.hash_to_bn(*crypto_items, params, _iteration=_iteration)
```
KPrasch: Hmm, very functional. :-) Is the `_iteration` parameter intended to be a use-case?  There is a need to encapsulate and recycle logic here, but exposing this type of input seems unusual to me (unless I'm missing something), since it's *only* used internally in this recursive call.  Perhaps this be composed as a closure or some other multi-function approach?

tuxxy: Ah yeah, that's a good idea:

`def hash_to_bn(*items, params):
    ...
    def iterate_hash(iteration=0)
        ...
```

KPrasch:
Also, not to start a naming-war of anything, but the title `hash_to_bn` is terse, non-descriptive, and misleading (not to mention it contains an acronym).  Is the intention that "hash" is meant to be used as a *verb*? If so, this is not obvious at all; rather this sounds like the function takes hashes as inputs!

michwill:
Hm.. To be fair, this is pretty unambiguous to me. If hash is a verb - that's obvious (bn will be the hash). If hash is a noun - that converts hash (of the argument) to bn, that's also correct

KPrasch:
@michwill a python double-entendre! I don't know that I agree.  Does this function takes *hashes* as inputs?

michwill:
It could certainly be understood as that also. But taking hashes as inputs and converting them to BN doesn't make more sense to me than universal conversion of bytes to BNs

umbral/curvebn.py
cygnusv: Yes, you're right. It's actually the first line. I like having everything in the try since it reads better in case there's not exception, which happens with overwhelming probability. But I recognize it's not a good practice, so I'll change it for a `try/except/else`.

PR #132, opened 2018-04-30
Using pre-compiled wheel.

jMyles:


michwill:
The PR worked for me, but only when doing `pipenv --dev --python 3.6` (because wheel is 3.6-specific)

Issue #123, opened 2018-04-19
KFrag.bn_id can be safely changed to a bitstring ID


cygnusv:
In previous versions of Umbral, we needed `KFrag.bn_id` to be a `BigNum`. This is not necessary anymore.

tuxxy:
It can effectively be a bytestring? What would the bytestring be of?

cygnusv:
I forgot to mention it should be a random bytestring, for example, of 32 bytes.

tuxxy:
Alright, presumably, it must be the size of the order of the curve used?

cygnusv:
Yes, since we use it to derive a `BigNum`, which has the order of the curve.

PR #134, opened 2018-05-03
[WIP] Improved Fragments

cygnusv:
Some improvements in `KFrag` and `CapsuleFrag`:
* Change `KFrag._bn_id` from a `CurveBN` type to a byte string (Solves #123), since it's not necessary anymore that this value be a `CurveBN`.
* New ephemeral public key in `KFrags` to use during reconstruction of activated capsule, instead of using Alice's public key. This new ephemeral pk is called `KFrag._point_seed_xcoord`. The motivation of this change is that if Alice's private key is disclosed, this prevents a potential reconstruction of the re-encryption key without using Bob's private key.
* Checks that all attached `cfrags` are consistent with respect to `point_noninteractive` and `point_seed_xcoord`
* Fixes corresponding tests

### DIFF
--- a/tests/test_capsule/test_capsule_serializers.py
+++ b/tests/test_capsule/test_capsule_serializers.py
@@ -3,6 +3,7 @@ import pytest
 from umbral import pre
 from umbral.curvebn import CurveBN
 from umbral.point import Point
+from umbral.config import default_curve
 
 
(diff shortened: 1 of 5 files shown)


cygnusv:
I preferred to do it through #140

PR #138, opened 2018-05-04
Remove loop in CurveBN.hash()

cygnusv:
Changes the try-and-increment method of the last stage of `CurveBN.hash()`, which converts a bytes digest into a valid `CurveBN`, to a direct method. Now, instead of computing `curvebn = digest % order` (which implies rejecting the case of `curvebn == 0` and trying again with an incremented input), we compute `curvebn = 1 + (digest % (order-1))`. This ensures that the result satisfies the condition `0 < curvebn < order`. With the change we get rid of the loop, the iteration counter, and handling exceptions. 

This is similar to the method described in Section B.2.1 of the [DSS Standard](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf) to produce per-message secrets in (EC)DSA.

### DIFF
--- a/umbral/curvebn.py
+++ b/umbral/curvebn.py
@@ -91,9 +91,10 @@ class CurveBN(object):
             try:
                 item_bytes = item.to_bytes()
             except AttributeError:
-                if not isinstance(item, bytes):
+                if isinstance(item, bytes):
+                    item_bytes = item
+                else:
                     raise TypeError("{} is not acceptable type, received {}".format(item, type(item)))
-                item_bytes = item
             blake2b.update(item_bytes)
 
         hash_digest = blake2b.finalize()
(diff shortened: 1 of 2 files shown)


umbral/curvebn.py
```diff
@@ -85,37 +85,36 @@ def hash(cls, *crypto_items, params=None):
...
-                    backend.openssl_assert(res == 1)
-                return cls(bignum, curve_nid, group, order)
-            except ValueError:
-                # This case is only reached when the result is 0, 
-                # which happens with prob. 1/order of the curve
-                iteration += 1
+                item_bytes = item.to_bytes()
+            except AttributeError:
+                if not isinstance(item, bytes):
+                    raise TypeError("{} is not acceptable type, received {}".format(item, type(item)))
```
jMyles: I don't really like raising in an `if not` instead of an `else`.  Also, this line is not tested.

PR #139, opened 2018-05-06
Proof and correctness logic reorganization

jMyles:
Bob needs a way to verify the correctness of individual cfrags.  This gives him that.

Also, I undertook a substantial rewrite of both `verify_correctness` and `_prove_correctness` in an effort to bring them closer to together, so that the reader may more easily understand how they inter-operate.

### DIFF
--- a/umbral/fragments.py
+++ b/umbral/fragments.py
@@ -232,8 +232,8 @@ class CapsuleFrag(object):
                        params: UmbralParameters=None):
 
         params = params if params is not None else default_params()
-        point_alice = pubkey_a.point_key
-        point_bob = pubkey_b.point_key
+        pubkey_a_point = pubkey_a.point_key
+        pubkey_b_point = pubkey_b.point_key
 
         ####
         ## Here are the formulaic constituents shared with `prove_correctness`.
@@ -266,8 +266,8 @@ class CapsuleFrag(object):
 
         g = params.g
 
-        g_y = (z2 * g) + (z1 * point_alice)
-        signature_input = [g_y, kfrag_id, point_alice, point_bob, u1, xcomp]
+        g_y = (z2 * g) + (z1 * pubkey_a_point)
+        signature_input = [g_y, kfrag_id, pubkey_a_point, pubkey_b_point, u1, xcomp]
         kfrag_signature1 = CurveBN.hash(*signature_input, params=params)
         valid_kfrag_signature = z1 == kfrag_signature1
 
--- a/umbral/fragments.py
+++ b/umbral/fragments.py
@@ -2,8 +2,10 @@ from cryptography.hazmat.primitives.asymmetric import ec
 
 from umbral.curvebn import CurveBN
 from umbral.config import default_curve, default_params
+from umbral.keys import UmbralPublicKey
 from umbral.point import Point
 from umbral.utils import get_curve_keysize_bytes
+from umbral.params import UmbralParameters
 
 from io import BytesIO
 
--- a/umbral/fragments.py
+++ b/umbral/fragments.py
@@ -1,5 +1,6 @@
 from cryptography.hazmat.primitives.asymmetric import ec
 
+from umbral._pre import assess_cfrag_correctness, verify_kfrag
 from umbral.curvebn import CurveBN
 from umbral.config import default_curve, default_params
 from umbral.keys import UmbralPublicKey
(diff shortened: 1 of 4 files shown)


umbral/pre.py
```diff
@@ -286,39 +287,44 @@ def reencrypt(kfrag: KFrag, capsule: Capsule, params: UmbralParameters=None,
 def _prove_correctness(cfrag: CapsuleFrag, kfrag: KFrag, capsule: Capsule, 
                        metadata: bytes=None, params: UmbralParameters=None
                       ) -> CorrectnessProof:
-    params = params if params is not None else default_params()
+    t = CurveBN.gen_rand(params.curve)
+    rk = kfrag._bn_key
 
-    e1 = cfrag._point_e1
-    v1 = cfrag._point_v1
+    params = params if params is not None else default_params()
```
cygnusv: `params` initialization should be done in the first line. I would also prefer to keep `t = CurveBN.gen_rand(params.curve)` just in the previous line to its first use `e2 = t * e`, but that's a matter of personal preference.

umbral/fragments.py
```diff
@@ -225,7 +227,61 @@ def to_bytes(self):
...
+    def verify_correctness(self, capsule: "Capsule",
+                       pubkey_a: UmbralPublicKey, pubkey_b: UmbralPublicKey,
+                       params: UmbralParameters=None):
+
+        params = params if params is not None else default_params()
+        pubkey_a_point = pubkey_a.point_key
+        pubkey_b_point = pubkey_b.point_key
+
+        ####
+        ## Here are the formulaic constituents shared with `prove_correctness`.
```
cygnusv: "... with `pre._prove_correctness`"

umbral/fragments.py
```diff
@@ -225,7 +227,61 @@ def to_bytes(self):
 
         return serialized_cfrag
 
-    def __bytes__(self):
-        return self.to_bytes()
+    def verify_correctness(self, capsule: "Capsule",
+                       pubkey_a: UmbralPublicKey, pubkey_b: UmbralPublicKey,
+                       params: UmbralParameters=None):
+
```
cygnusv: I must say that I prefer to keep the logic in `umbral.pre` and to have this only as a wrapper (didn't we agree on Discord to wrap this?). Now that I think about it, it's the same for `KFrag.verify`.

umbral/pre.py
```diff
@@ -252,7 +252,8 @@ def split_rekey(priv_a: Union[UmbralPrivateKey, CurveBN],
         u1 = rk * u
         y = CurveBN.gen_rand(params.curve)
 
-        z1 = CurveBN.hash(y * g, id_kfrag, pub_a, pub_b, u1, xcomp, params=params)
+        signature_input = [y * g, id_kfrag, pubkey_a_point, pubkey_b_point, u1, xcomp]
+        z1 = CurveBN.hash(*signature_input, params=params)
         z2 = y - priv_a * z1
```
cygnusv: Same TODO above this block

umbral/fragments.py
tuxxy: Yeah, I think I prefer this too. Just to keep the logic where the rest of the cryptographic code is.

@cygnusv Am I right that all of the logic described in the Umbral paper is in `umbral.pre`?

jMyles: Ahh, right, we did indeed agree to wrap this - that slipped my mind.

As for keeping all the "cryptographic" code together: it's far from obvious how to do that.  There's code reasonable understood as "cryptography" in every single module, including this one.  

But as for wrapping this, I'll mark this [WIP] and get a fix in later today.  :-)

umbral/fragments.py
jMyles: :+1:

umbral/pre.py
jMyles: Right - `params` was a goof here.  

On `t`, I had the same inclination, but then that makes it impossible to do a line-for-line match with the verify logic.

umbral/fragments.py
cygnusv: Well, I was talking about Umbral KEM itself. Of course there's cryptographic code in `umbral.point` or `umbral.openssl`, but that's a different abstraction layer, in particular, a lower level. If we move the logic of `CFrag.verify_correctness` and `KFrag.verify` to `umbral.pre`, then I think all the core cryptographic logic of Umbral is in `umbral.pre`, in approximately 600 lines of code. That's neat, IMHO.

cygnusv:
@tuxxy Not all, but most of it. What parts of Umbral are not in `umbral.pre`?
- Key generation is in `umbral.keys`, but that's reasonable since Umbral encryption keys follow the same simple approach than most schemes based on ElGamal/Diffie-Hellman (i.e., `pk = sk * g).
- Some auxiliary stuff in `umbral.util`, which is basically non-cryptographic code.

PR #140, opened 2018-05-08
Improved Fragments

cygnusv:
Some improvements in `KFrag` and `CapsuleFrag`:
* Change `KFrag._bn_id` from a `CurveBN` type to a byte string (Solves #123), since it's not necessary anymore that this value be a `CurveBN`.
* New ephemeral public key in `KFrags` to use during reconstruction of activated capsule, instead of using Alice's public key. This new ephemeral pk is called `KFrag._point_xcoord`. The motivation of this change is that if Alice's private key is disclosed, this prevents a potential reconstruction of the re-encryption key without using Bob's private key.
* Checks that all attached `cfrags` are consistent with respect to `point_noninteractive` and `point_xcoord`
* Fixes corresponding tests

### DIFF
--- a/tests/test_keys/test_key_fragments.py
+++ b/tests/test_keys/test_key_fragments.py
@@ -1,6 +1,6 @@
 from umbral import pre
-import time
-
+from umbral.config import default_curve
+from umbral.fragments import KFrag, CapsuleFrag
 
 def test_kfrag_serialization(alices_keys):
     priv_key_alice, pub_key_alice = alices_keys
@@ -8,14 +8,15 @@ def test_kfrag_serialization(alices_keys):
     kfrags = pre.split_rekey(priv_key_alice, pub_key_alice, 1, 2)
     kfrag_bytes = kfrags[0].to_bytes()
 
-    # A KFrag can be represented as the 194 total bytes of two Points (33 each) and four CurveBNs (32 each).
-    assert len(kfrag_bytes) == 33 + 33 + (32 * 4) == 194
+    curve = default_curve()
+    assert len(kfrag_bytes) == KFrag.get_size(curve)
 
     new_frag = pre.KFrag.from_bytes(kfrag_bytes)
     assert new_frag._id == kfrags[0]._id
     assert new_frag._bn_key == kfrags[0]._bn_key
     assert new_frag._point_noninteractive == kfrags[0]._point_noninteractive
     assert new_frag._point_commitment == kfrags[0]._point_commitment
+    assert new_frag._point_xcoord == kfrags[0]._point_xcoord
     assert new_frag._bn_sig1 == kfrags[0]._bn_sig1
     assert new_frag._bn_sig2 == kfrags[0]._bn_sig2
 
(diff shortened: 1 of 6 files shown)


tests/test_keys/test_key_fragments.py
```diff
@@ -27,11 +28,7 @@ def test_cfrag_serialization_with_proof_and_metadata(alices_keys):
     kfrags = pre.split_rekey(priv_key_alice, pub_key_alice, 1, 2)
 
     # Example of potential metadata to describe the re-encryption request
-    metadata = { 'ursula_id' : 42, 
-                 'timestamp' : time.time(), 
-                 'capsule' : bytes(capsule), 
-               }
-    metadata = str(metadata).encode()
+    metadata = b'This is an example of metadata for re-encryption request'
```
tuxxy: Is there anything more useful we can put in here that may demonstrate its utility? This can be NuCypher agnostic, too.

umbral/_pre.py
```diff
@@ -112,19 +114,22 @@ def verify_kfrag(kfrag,
...
     #  We check that the commitment u1 is well-formed
     correct_commitment = u1 == key * u
 
     # We check the Schnorr signature over the kfrag components
     g_y = (z2 * params.g) + (z1 * pubkey_a_point)
 
-    kfrag_components = [g_y, kfrag._bn_id, pubkey_a_point, pubkey_b_point, u1, x]
-    valid_kfrag_signature = z1 == CurveBN.hash(*kfrag_components, params=params)
+    signature_input = [g_y, id, pubkey_a_point, pubkey_b_point, u1, ni, xcoord]
+    valid_kfrag_signature = z1 == CurveBN.hash(*signature_input, params=params)
```
tuxxy: You can actually just do:

`valid_kfrag_signature = z1 == CurveBN.hash(g_y, id, pubkey_a_point, pubkey_b_point, u1, ni, xcoord, params=params)`

This avoids making the list first.

umbral/pre.py
```diff
@@ -216,50 +224,66 @@ def split_rekey(priv_a: Union[UmbralPrivateKey, CurveBN],
...
-    blake2b.update(g_ab.to_bytes())
+    blake2b.update(dh_xcoord.to_bytes())
     hashed_dh_tuple = blake2b.finalize()
 
+    bn_size = CurveBN.get_size(params.curve)
+
     kfrags = []
     for _ in range(N):
-        id_kfrag = CurveBN.gen_rand(params.curve)
+        id = os.urandom(bn_size)
```
tuxxy: Yay! +1 for nondeterminism!

tests/test_keys/test_key_fragments.py
cygnusv: Sure, but we went over this before and we ended up with this again.

tuxxy: But maybe we can demonstrate using a timestamp or  perhaps even the capsule itself?

umbral/_pre.py
cygnusv: I know, but the list is becoming a little bit long, plus I wanted to stress what's what we are signing.

tests/test_keys/test_key_fragments.py
cygnusv: Yes, I did that in #108 but it was decided in #115 to revert to the simple approach.

See https://github.com/nucypher/pyUmbral/pull/115#discussion_r183594166

umbral/_pre.py
jMyles: Yeah, well I also wondered why we were making it a list here - it never mutates.  Why not a tuple?

umbral/pre.py
jMyles: I guess I'm missing something here - how is this less deterministic than `BN_rand_range`?

tuxxy: Oh, I'm just putting a general, positive comment on a PR. Nothing special. This makes the ID a bitstring and I think it's great we can use `os.urandom` here.

PR #141, opened 2018-05-14
[WIP] Add some changes for Point validation

tuxxy:
### What this does:
1. Adds `salt` and `info` params to `utils.kdf`.
2. Changes `openssl._bn_is_on_curve` to `openssl._bn_is_within_order`.
3. Adds `openssl._is_point_on_curve` function.
4. Checks that a `Point` is on a curve.
5. Checks that a given `UmbralPublicKey` is on a curve.

### DIFF
--- a/umbral/openssl.py
+++ b/umbral/openssl.py
@@ -148,6 +148,24 @@ def _get_affine_coords_via_EC_POINT(ec_point, ec_group=None, curve_nid: int=None
     return (affine_x, affine_y)
 
 
+def _is_point_on_curve(ec_point, ec_group=None, curve_nid: int=None):
+    """
+    Returns a boolean value for whether or not a given ec_point is on the
+    provided curve_nid.
+    """
+    if curve_nid:
+        ec_group = _get_ec_group_by_curve_nid(curve_nid)
+    elif not ec_group:
+        raise ValueError("No group provided.")
+
+    with backend._tmp_bn_ctx() as bn_ctx:
+        res = backend._lib.EC_POINT_is_on_curve(ec_group, ec_point, bn_ctx)
+        if res <= 0:
+            return False
+        elif res == 1:
+            return True
+
+
 @contextmanager
 def _tmp_bn_mont_ctx(modulus):
     """
--- a/umbral/openssl.py
+++ b/umbral/openssl.py
@@ -160,11 +160,7 @@ def _is_point_on_curve(ec_point, ec_group=None, curve_nid: int=None):
 
     with backend._tmp_bn_ctx() as bn_ctx:
         res = backend._lib.EC_POINT_is_on_curve(ec_group, ec_point, bn_ctx)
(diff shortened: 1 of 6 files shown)


umbral/openssl.py
```diff
@@ -148,6 +148,24 @@ def _get_affine_coords_via_EC_POINT(ec_point, ec_group=None, curve_nid: int=None
...
+    elif not ec_group:
+        raise ValueError("No group provided.")
+
+    with backend._tmp_bn_ctx() as bn_ctx:
+        res = backend._lib.EC_POINT_is_on_curve(ec_group, ec_point, bn_ctx)
+        if res <= 0:
+            return False
+        elif res == 1:
+            return True
+
```
cygnusv: > "_EC_POINT_is_on_curve returns 1 if the point is on the curve, 0 if not, or -1 on error._"

I see two options here, depending if we want to handle the case of an error or not:
- If we want to handle it, change the block to something like:
```
if res > 0:
    return res == 1
else:
    raise ValueError("Wrong input")
```
- In case we don't want to handle the error, just change the `if/elif` block to ` return res == 1`

umbral/utils.py
```diff
@@ -28,14 +28,14 @@ def poly_eval(coeff, x):
     return result
 
 
-def kdf(ecpoint, key_length):
+def kdf(ecpoint, key_length, salt=None, info=None):
```
cygnusv: I'm thinking this should be called `derive_key_from_point` since the input is presumed to be a point. Hmmm, perhaps this belongs to `umbral.Point`?

umbral/openssl.py
```diff
@@ -54,7 +54,7 @@ def _get_ec_generator_by_curve_nid(curve_nid: int):
     return generator
 
 
-def _bn_is_on_curve(check_bn, curve_nid: int):
+def _bn_is_within_order(check_bn, curve_nid: int):
```
cygnusv: Ah, great name, much more informative :)

umbral/point.py
```diff
@@ -264,6 +266,9 @@ def unsafe_hash_to_point(data, params, label=None):
             else:
                 # Any other exception, we raise it
                 raise e
+        except ValueError as e:
```
cygnusv: We should catch the `InternalErrors` above within `Point.from_bytes` and raise a unified `ValueError` instead, and catch that here.

michwill:
The segfault fix (hopefully)
https://github.com/tuxxy/pyUmbral/pull/2

umbral/utils.py
jMyles: Hmmm, yeah that seems reasonable.

KPrasch:
So - this issue has grown beyond the scope a WIP? Perhaps we can either close this PR or re-label it?

tuxxy:
We can go ahead and close it. :)

PR #143, opened 2018-05-18
Using getrandom(...) instead of os.urandom(...) where it makes sense.

jMyles:
There are a couple of places where it makes sense to use the stronger `getrandom` than `os.urandom`:

* When generating keying material, I think it's reasonable to imagine that users will not be blocked by a blocking system random, as this call is likely to be made only rarely.

* When generating KFrags, it's probably reasonable to imagine that users will adopt a slightly higher level of paranoia even if it means a minor inconvenience.  

Additionally, these are places where it seems that a highest quality random source is paramount.

### DIFF
--- a/umbral/keys.py
+++ b/umbral/keys.py
@@ -275,7 +275,7 @@ class UmbralKeyingMaterial(object):
                 raise ValueError("UmbralKeyingMaterial must have size at least 32 bytes.")
             self.keying_material = keying_material
         else:
-            self.keying_material = os.urandom(64)
+            self.keying_material = os.getrandom(32, flags=1)
 
     def derive_privkey_by_label(self, label: bytes, salt: bytes=None, 
                                 params: UmbralParameters=None):
--- a/umbral/pre.py
+++ b/umbral/pre.py
@@ -266,7 +266,7 @@ def split_rekey(privkey_a_bn: Union[UmbralPrivateKey, CurveBN],
 
     kfrags = []
     for _ in range(N):
-        id = os.urandom(bn_size)
+        id = os.getrandom(bn_size, flags=1)
 
         share_x = CurveBN.hash(id, hashed_dh_tuple, params=params)
 


jMyles:
> I have no idea what `getrandom` does differently from `os.urandom`

It's not too late to get with the program ;-)

https://www.systutorials.com/docs/linux/man/2-getrandom/

> there is an idiom that says, "always use /dev/urandom, not /dev/random."

...an idiom?

I don't find your link at all convincing.  Indeed, /dev/random is plainly superior.  I mean, it's a small enough distinction as to not make too much of a distinction, so you know, I'm find with urandom.  But if we can swing random, that's better.

> There is no security benefit from using /dev/random

Sure there is.  It's the better option where blocking is an option.  I'm surprised you're disputing that.

> entropy collection timing side channels. 

Eh?  I have literally never heard of this.  I'll need a link.

tuxxy:
No, it's not better. Go find one of the Twitter flame wars on this discussion and this paper: 
http://eprint.iacr.org/2013/338.pdf.

You're falling prey to the many misunderstood views.

jMyles:
Dude.  Seriously?

Can you give me a link to "entropy collection timing side channels"?  This phrase does not appear on google.  I've literally never heard of this.  I'm genuinely interested, and surprised to hear of this phenomenon.  If it's real, I really, honestly, want to read about it, so link a brother.

But anyhow, short of such an exotic attack, I assure you, I'm not "falling prey" to a misunderstood view.  

In situations where blocking is acceptable, gathering entropy is worth blocking for.  In situations where blocking is going to cause users to find workarounds, then yeah let's not block.  Obviously.  I've been using `/dev/urandom` my whole life because I've been working on web stuff.

In our case, we have some cases where blocking is probably unacceptable.  As you can see, I left those alone.  In cases where blocking is OK... well, blocking is OK.

I have exactly zero interest in twitter flame wars.  Plenty of sober, serious inventors are turning onto the need for a more cognizable API for blocking random, hence the addition of `getrandom` to the linux kernel, python core library, and everywhere else.  I just linked you; go check it out.

tuxxy:
By blocking until you have enough entropy it inherently opens itself up to timing side channels. Here's a GitHub issue in Libsodium  that brings up the same things: https://github.com/jedisct1/libsodium/issues/49.

I have exactly zero interest in having yet-another-discussion-on-random-vs-urandom because of the flame topic that it is.

The entropy from each is _the exact same_. Did you really read the myths page? Because this random vs urandom is an age old misunderstood argument built on paranoia for no reason.

tuxxy:
In fact, before I forget, the entirety of cryptography.io is built on the /dev/urandom backend for entropy. They hacked OpenSSL to use this instead of OpenSSL's userspace CSPRNG because /dev/urandom is better.

jMyles:
> The entropy from each is the exact same. 

In some cases, this is true.  In some (important) cases, it's not.  Why ignore those?  They're actually plausible for our purposes.

This second link you've provided links to your paper as its only source for its only comment.  And it's not at all convincing.  I can't tell if the paper is suffering from a language barrier or just melodrama, but it repeatedly says that neither source are robust, but doesn't providing convincing evidence that the attacks is describes are actually possible in any real-world scenario I can imagine.

What is your theory about `getrandom` then?  That it's entirely built on "an age old misunderstood argument built on paranoia for no reason."  ?? Haha, I assure you it's not.

> In fact, before I forget, the entirety of cryptography.io is built on the /dev/urandom backend for entropy. They hacked OpenSSL to use this instead of OpenSSL's userspace CSPRNG because /dev/urandom is better.

So, just to be clear: you understand the advantage of urandom, right?  I think then you'd understand why they made the reasonable choice (in fact, the same choice as the python stdlib, until recently) to give users `urandom`, which is a more usable, reliably performant interface.  Of course!

But now, in 2018, is it the best for us?  As tooling is being built against *unknown* attacks - and let's be clear: it's plausible (albeit unlikely) that there exists an undisclosed or classified attack against `urandom` but not `/dev/random` - what is the sense of refusing to use it?

I don't want to hear about twitter arguments; I want to hear a real reason to resist this change today. 

If you're saying that `/dev/urandom` is a good solution in almost every case ever, then we're in agreement.  I think everybody agrees with this. I love urandom, man!  We all do.

If you're saying that blocking random is a bad idea tout court with no good use cases ever, then we don't agree (and in fact, I don't think that there exists a mainstream view for that position today).

Now, do we *need* Wthe more pure solution?  Of course not.  Look: I closed this PR.  I'm done with it.  But at the end of the day, is using a blocking source a good solution for these two calls?  ...yeah, it is.  When we come to stop supporting python versions and linux kernels that lack the simple, new, well-researched interface for this purpose, then I think it's a good idea for us to just do it.

jMyles:
It's kinda crazy to me that you're bringing this up in the context of a closed PR - you obviously feel strongly about this.

I'm totally fine with /dev/urandom, just as you are.  I think everybody is.

But I was reminded at PyCon that getrandom is now a thing, and I thought, "hey, yeah, there are a couple places where we can afford to be super duper purists."

At the end of the day, urandom is great and fills a huge need, and we benefit from it.  I'm not saying not to use it; I just think that it behooves us to use the best matches for our needs as they become available.  Right now, that appears to be /dev/urandom, which is why I closed this PR.

michwill:
Yeah, this was at times a very flame-generating topic indeed.
But I suspect that when we want secure keys, we can xor the random numbers with hashes of webcam stream :-D
Otherwise, looks like cryptographic community is very comfortable with urandom

jMyles:
It turns out that this is a non-issue for python 3.6, as `os.urandom` does in fact use `getrandom` under the hood.

jMyles:
For more information about the underlying issue, see this question that I opened on the security stackexchange: https://security.stackexchange.com/questions/186086/generating-key-material-shortly-after-boot-in-a-vm-a-use-case-for-blocking-inst/186097#186097

jMyles:
And @michwill, to directly respond to your observation: I think that because of edge cases like this one, the "cryptographic community" (to the degree that that's a cognizable subset of humanity) *isn't* entirely comfortable with `/dev/urandom` - that's why `getrandom` has been introduced and is becoming more normal.   That's the whole point I was making by opening to PR (although I now realize that it wasn't necessary, as `os.urandom` no longer directly calls `/dev/urandom`).

PR #149, opened 2018-05-25
Improved Point deserialization

cygnusv:
- Introduces `get_ec_group_degree`, a new method in `umbral.openssl` to compute the expected size of serialized Points based on the size of the order of the underlying finite field. Previously we were using the size of the group order, which usually it's the same size, but not necessarily. 
- Improved checks in `Point.from_bytes`: now only bytestrings with the exact possible sizes are accepted.
- Tests for malformed input byte strings to `Point.from_bytes`.
- Fixes the problem in `test_point_serializers.py/test_bytes_serializers`. Previously it was marked as `xfail`. Now all tests in the codebase are a success (hooray! 🎉)

### DIFF
--- a/umbral/utils.py
+++ b/umbral/utils.py
@@ -3,7 +3,8 @@ import math
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
-
+from cryptography.hazmat.primitives.asymmetric import ec
+from umbral import openssl
 
 def lambda_coeff(id_i, selected_ids):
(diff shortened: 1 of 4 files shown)


umbral/utils.py
```diff
@@ -42,3 +43,16 @@ def kdf(ecpoint, key_length):
 
 def get_curve_keysize_bytes(curve):
     return (curve.key_size + 7) // 8
+
+def get_field_order_size_in_bytes(curve):
+
```
tuxxy: Only thing I would change is to drop the 'in' so it's `get_field_order_size_bytes`, but this looks great!

PR #150, opened 2018-05-26
expected_bytes_length

jMyles:


tuxxy:
I don't see a benefit to `expected_bytes_length` over `get_size`. The latter is shorter and more concise and we already use it in many places.

jMyles:
Here's a brief discussion we had about `get_size`:

---
[10:47 AM] jMyles: 1) I don't think get_size is a good name.  For example, for BigNums, don't you expect get_size to be an instance method that helps you determine which number is bigger?  The method doesn't really tell you anything about the size of the object, just about the length of the bytestring you can expect to represent the object (regardless of its size)
[10:47 AM] jMyles: 2) A classmethod is probably better than a class attribute, so I do like that part of the pyUmbral convention better.
[10:48 AM] michwill: get_bytesize?
[10:48 AM] jMyles: (or, as @michwill points out, it can even be a staticmethod in some cases)
[10:48 AM] jMyles: get_bytes_length?
1
[10:49 AM] jMyles: Also: I am always a little wary of having externally-enforced opinions on the method name for a class.  Remember: bytestringSplitter can also use classes over which we don't have control.
[10:51 AM] michwill: Hmm, what do you mean by externally-enforced opinions?
[11:04 AM] jMyles: Well, by saying get_bytes_length, we assume that any object we'd ever pass to bytestringSplitter (without manually specifying a length) doesn't use that name for something else.
[11:06 AM] michwill: Could be with underscore 
[11:06 AM] michwill: Of course, could be __len__, but I can see how it can go wrong
[11:10 AM] jMyles: I don't think there is a way to make a magic method for len for a class object, but if there were, I think this is exactly what it oughta be for 
[11:11 AM] jMyles: (actually... what happens if you define __len__ on a metaclass?)
[11:11 AM] jMyles: (not that that's the right way to do this, per se)
[11:13 AM] michwill: Ah yeah, true, you want len of a class rather than an object
[11:21 AM] jMyles: Actually, that does appear to work, although it's undocumented.
[11:25 AM] michwill: What exactly appears to work? __len__ on a metaclass IMO should work: that's very logical
[11:25 AM] jMyles: yep, agreed
[11:25 AM] jMyles: I've just never seen it, nor can I find documentation of it
[11:25 AM] michwill: But obviously we don't want the developer to specify metaclass every time, right?
[11:26 AM] michwill: I guess, we want to do something like __len_bytes() or whatever, which gets called by the metaclass' __len__
[11:34 AM] jMyles: OK
[11:35 AM] jMyles: So, the metaclass thing is definitely interesting, but probably the wrong approach for us in pyUmbral / nucypher because the size is dependant on the curve, and the user doesn't have to have specified the curve (or, in the case of nucypher, won't have specified the curve) by the time the class is constructed from the metaclass.
[11:35 AM] jMyles: So yeah, I'm back to thinking that a regular classmethod is best
[11:35 AM] jMyles: and let's just pick a name and be done with it.
[11:36 AM] jMyles: I'm fine with bytes_length or get_bytes_length or whatever.
[11:37 AM] jMyles: ...although again, bytes_length sounds like an instance method - "how long will this object be in bytes?"
[11:37 AM] jMyles: whereas, something like, expected_bytes_length conveys that it is the length we expect the representation of an instance to be(edited)
[11:38 AM] michwill: I'm ok with expected. Should we have it underscored, or not really?
[11:38 AM] jMyles: Probably not - it is certainly intended to be public.

PR #157, opened 2018-05-30
Attaching CFrags to a Capsule now requires correctness check and KFrag signature validation.

jMyles:


umbral/pre.py
```diff
@@ -108,6 +115,83 @@ def from_bytes(cls, capsule_bytes: bytes, curve: ec.EllipticCurve = None):
         components = splitter(capsule_bytes)
         return cls(*components)
 
+    def get_or_set_delegating_key(self, delegating_key: UmbralPublicKey = None):
```
michwill: Hm.. I wonder if it can be done through `__getattr__` / `__setattr__`. Of course, that'd be too low-level (and, hence, error-prone). So, I wonder if Python has something to construct a "set-once" property conveniently

michwill: Like, using a setter:
```
In [1]: class X(object):
   ...:     def __init__(self):
   ...:         self.__x = None
   ...:     @property
   ...:     def x(self):
   ...:         return self.__x
   ...:     @x.setter
   ...:     def x(self, value):
   ...:         if self.__x is not None:
   ...:             raise ValueError('oops')
   ...:         else:
   ...:             self.__x = value
   ...:             

In [2]: y = X()

In [3]: y.x

In [4]: y.x = 1

In [5]: y.x
Out[5]: 1

In [6]: y.x = 1
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-183902452f08> in <module>()
----> 1 y.x = 1

<ipython-input-1-118622fe2a80> in x(self, value)
      8     def x(self, value):
      9         if self.__x is not None:
---> 10             raise ValueError('oops')
     11         else:
     12             self.__x = value

ValueError: oops
```

jMyles: Yeah, I thought about it - I can do that instead if you prefer.

tuxxy: When I had the idea for `get_or_set`, I meant it to work for the model I proposed. I don't really like it here, tbh.

michwill: I think, that'd be more readable where it's used.
Yeah, also what feel strange also is that we repeat _exactly the same thing_ multiple times

jMyles: Heh, yeah.  When I first wrote it, I was thinking it was only going to be once.  Then twice.  Then I realize it needed to be three times.

jMyles: OK.  I'll rework it then.

tests/test_capsule/test_capsule_correctness_checks.py
```diff
@@ -0,0 +1,157 @@
...
+    We need the proper keys to verify the correctness of CFrags
+    in order to attach them to a Capsule.
+    """
+    capsule = Capsule(point_e=Point.gen_rand(),
+                      point_v=Point.gen_rand(),
+                      bn_sig=CurveBN.gen_rand())
+
+    cfrag = CapsuleFrag(point_e1=Point.gen_rand(),
+                        point_v1=Point.gen_rand(),
+                        kfrag_id=os.urandom(10),
```
tuxxy: This needs to be 32 bytes, not 10.

umbral/pre.py
```diff
@@ -108,6 +115,40 @@ def from_bytes(cls, capsule_bytes: bytes, curve: ec.EllipticCurve = None):
...
+    def set_delegating_key(self, key: UmbralPublicKey):
+        return self._set_cfrag_checking_key("delegating", key)
+
+    def set_encrypting_key(self, key: UmbralPublicKey):
+        return self._set_cfrag_checking_key("encrypting", key)
+
+    def set_verifying_key(self, key: UmbralPublicKey):
+        return self._set_cfrag_checking_key("verifying", key)
+
+    def set_three_keys(self,
```
tuxxy: With the removal of each singular key setting method, I think that `set_verification_keys` is better and more clear, in the name, with what is going on.

umbral/pre.py
```diff
@@ -130,9 +171,30 @@ def verify(self, params: UmbralParameters = None) -> bool:
...
+                     params: UmbralParameters = None) -> None:
+
+        delegating_key_is_new = self.set_delegating_key(delegating_pubkey)
+        encrypting_key_is_new = self.set_encrypting_key(encrypting_pubkey)
+        verifying_key_is_new = self.set_verifying_key(verifying_pubkey)
+
+        self.verify_cfrag(cfrag, params)
         self._attached_cfrags.append(cfrag)
 
+        return delegating_key_is_new, encrypting_key_is_new, verifying_key_is_new
```
tuxxy: I don't like the idea of a method with the name `attach` returning _anything_. It's pretty obvious what the user is providing, I don't think they need to have it returned to them.

umbral/pre.py
```diff
@@ -108,6 +115,40 @@ def from_bytes(cls, capsule_bytes: bytes, curve: ec.EllipticCurve = None):
...
+    def set_three_keys(self,
+                              delegating_key: UmbralPublicKey = None,
+                              encrypting_key: UmbralPublicKey = None,
+                              verifying_key: UmbralPublicKey = None
+                              ):
+        delegating_key_details = self.set_delegating_key(delegating_key)
+        encrypting_key_details = self.set_encrypting_key(encrypting_key)
+        verifying_key_details = self.set_verifying_key(verifying_key)
+
+        return delegating_key_details, encrypting_key_details, verifying_key_details
```
tuxxy: Again, I find it pretty redundant that a `set` method returns what the user just provided?

tests/test_capsule/test_capsule_correctness_checks.py
```diff
@@ -0,0 +1,157 @@
...
+    However, even when properly attaching keys, we can't attach the CFrag
+    if it is unproven.
+    """
+    capsule = Capsule(point_e=Point.gen_rand(),
+                      point_v=Point.gen_rand(),
+                      bn_sig=CurveBN.gen_rand())
+
+    cfrag = CapsuleFrag(point_e1=Point.gen_rand(),
+                        point_v1=Point.gen_rand(),
+                        kfrag_id=os.urandom(10),
```
tuxxy: Also needs to be 32.

jMyles: ...why?

umbral/pre.py
jMyles: It doesn't anymore - it just returns `True` if it was set new.

tests/test_capsule/test_capsule_correctness_checks.py
jMyles: Why?  What's so special about 32?  Just that it happens to be the size of a bignum represented as bytes given a curve that we use by default?  There's no harm in making it some other arbitrary number in a test.

tuxxy: Because it's the way it should be (for 256-bit curves) and if there is a _right_ way to do it, then it should be done that way, imho.

jMyles: ...why is that the right way?  Who says kfrag IDs have to follow curves at all?  And if they do, why 256-bit curves?  I mean, I can change it to 32, but I don't think these are good arguments.  The size is arbitrary, and I don't want to give the impression that 32 bytes is the only possible length either.

umbral/pre.py
jMyles: Eh?  They might honestly not know if they have set a new key or not.  Why not tell them?

tuxxy: I think I would prefer to keep this out until we know a clear need for something like this.

tests/test_capsule/test_capsule_correctness_checks.py
tuxxy: By convention, the ID should be as long as the curve bitsize

jMyles: Yeah?  What convention?  Is that documented somewhere?

It's just a test.  There's really no reason for it to even be random in the first place.  It can just be `b"this is a kfrag ig"`

tests/test_simple_api.py
```diff
@@ -27,13 +27,17 @@ def test_simple_api(alices_keys, N, M, curve=default_curve()):
...
+                     verifying=signing_privkey.get_pubkey())
+
     cleartext = pre.decrypt(ciphertext, capsule, delegating_privkey, params=params)
     assert cleartext == plain_data
 
     kfrags = pre.split_rekey(delegating_privkey, signer, decrypting_key.get_pubkey(), M, N, params=params)
     for kfrag in kfrags:
         cfrag = pre.reencrypt(kfrag, capsule, params=params)
-        capsule.attach_cfrag(cfrag)
+        capsule.attach_cfrag(cfrag, params=params)
```
michwill: Maybe not for this PR, but.. Why do we have to carry the params? Cannot the params be set when the capsule is originally instantiated?

jMyles: Yeah, that occurred to me.  There's... probably no good reason.  When I started working on this PR, I was in a way of trying to be sort of minimally invasive with respect to these sorts of changes, but that appears to have been a bit misguided.  :-)

PR #158, opened 2018-05-31
Renaming keypairs

cygnusv:
## What this does ##

- Renaming of keypairs and associated keys to follow the new standard (delegating, receiving, verifying)
- Fixes some weird test patterns, where `pre.split_rekey` is used from Alice to Alice.
- Inserts a check in `Capsule._set_cfrag_correctness_key` to allow only public keys associated to delegating, receiving, or verifying.
- Fixes a bug in `UmbralPrivateKey`, in the generation of the associated public key, which was created with the default params. This produced problems when used with curves other than default.

### DIFF
--- a/umbral/keys.py
+++ b/umbral/keys.py
@@ -29,7 +29,7 @@ class UmbralPrivateKey(object):
 
         self.params = params
         self.bn_key = bn_key
-        self.pubkey = UmbralPublicKey(self.bn_key * self.params.g)
+        self.pubkey = UmbralPublicKey(self.bn_key * self.params.g, params=params)
 
     @classmethod
     def gen_key(cls, params: UmbralParameters=None):
--- a/umbral/pre.py
+++ b/umbral/pre.py
@@ -119,18 +119,21 @@ class Capsule(object):
         return cls(*components)
 
     def _set_cfrag_correctness_key(self, key_type, key: UmbralPublicKey):
+        if key_type not in ("delegating", "receiving", "verifying"): 
+            raise ValueError("You can only set 'delegating', 'receiving' or 'verifying' keys.") 
+
         current_key = self._cfrag_correctness_keys[key_type]
 
         if current_key is None:
             if key is None:
-                raise TypeError("The Delegating Key is not set and you didn't pass one.")
+                raise TypeError("The {} key is not set and you didn't pass one.".format(key_type))
             else:
                 self._cfrag_correctness_keys[key_type] = key
                 return True
-        elif key in (None, self._cfrag_correctness_keys[key_type]):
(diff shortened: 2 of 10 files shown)


umbral/pre.py
```diff
@@ -119,29 +119,32 @@ def from_bytes(cls, capsule_bytes: bytes, curve: ec.EllipticCurve = None):
...
     def _set_cfrag_correctness_key(self, key_type, key: UmbralPublicKey):
+        if key_type not in ("delegating", "receiving", "verifying"): 
+            raise ValueError("You can only set 'delegating', 'receiving' or 'verifying' keys.") 
+
         current_key = self._cfrag_correctness_keys[key_type]
 
         if current_key is None:
             if key is None:
-                raise TypeError("The Delegating Key is not set and you didn't pass one.")
+                raise TypeError("The {} key is not set and you didn't pass one.".format(key_type))
```
jMyles: Ahh, nice catch.

umbral/pre.py
```diff
@@ -119,29 +119,32 @@ def from_bytes(cls, capsule_bytes: bytes, curve: ec.EllipticCurve = None):
...
             else:
                 self._cfrag_correctness_keys[key_type] = key
                 return True
-        elif key in (None, self._cfrag_correctness_keys[key_type]):
+        elif key in (None, current_key):
```
jMyles: Right right.  That was sloppalicious on my part.

umbral/keys.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, bn_key: CurveBN, params: UmbralParameters=None):
 
         self.params = params
         self.bn_key = bn_key
-        self.pubkey = UmbralPublicKey(self.bn_key * self.params.g)
+        self.pubkey = UmbralPublicKey(self.bn_key * self.params.g, params=params)
```
michwill: Btw do we ever instantiate `UmbralPublicKey` directly? If not, should we make `params` a required argument?

jMyles: We do frequently instantiate `UmbralPublicKey` directly and frequently - by deserializing.

michwill: I wouldn't call it "directly" - that's deserialization which gets called by the user. E.g. `UmbralPublicKey.from_bytes`, not `__init__` directly.
My point is that all methods which are not to be called by the user directly can make `params` a required argument, so that we don't forget to set it accidentally

jMyles: Haha, right.  It's almost like `__init__` is private.  ;-)

cygnusv: Yeah, good comment. I already took care of making params a required argument in the next PR.

PR #159, opened 2018-05-31
Simplified API

cygnusv:
## What this does: ##
This PR provides a slightly simplified API for pyUmbral. In particular, removes all redundant arguments from public and private methods that could be obtained by other means (primarily, from a neighboring argument), such as `params` and correctness keys. The result is not only that the interface is simplified, but also that there is less surface for misuse. 

* Simplified public API for `umbral.pre` and `umbral.fragments`:
    * Methods now only take Umbral keys as arguments, rather than primitive types (Point, CurveBN).
    * Remove unnecessary arguments from public facing and internal methods when they can be extracted from a `Capsule`, `UmbralPublicKey` or `UmbralPrivateKey`.
    * Adapts the test suite to new simplified API
* Fixes a bug in `Capsule.attach_cfrag()` that allowed to attach incorrect `CFrag`
* Adds a getter in `Capsule` for correctness keys
* Removes `pre.CHACHA20_KEY_SIZE` constant and uses `dem.DEM_KEYSIZE` instead
* Add `__eq__` method to `UmbralParameters`

### DIFF
--- a/tests/test_capsule/test_capsule_operations.py
+++ b/tests/test_capsule/test_capsule_operations.py
@@ -62,11 +62,11 @@ def test_decapsulation_by_alice(alices_keys):
 
     delegating_privkey, _signing_privkey = alices_keys
 
-    sym_key, capsule = pre._encapsulate(delegating_privkey.get_pubkey().point_key, params)
+    sym_key, capsule = pre._encapsulate(delegating_privkey.get_pubkey())
     assert len(sym_key) == 32
 
     # The symmetric key sym_key is perhaps used for block cipher here in a real-world scenario.
-    sym_key_2 = pre._decapsulate_original(delegating_privkey.bn_key, capsule)
+    sym_key_2 = pre._decapsulate_original(delegating_privkey, capsule)
     assert sym_key_2 == sym_key
 
 
@@ -115,8 +115,7 @@ def test_capsule_as_dict_key(alices_keys, bobs_keys):
     capsule.attach_cfrag(cfrag)
 
     # Even if we activate the capsule, it still serves as the same key.
-    cleartext = pre.decrypt(ciphertext, capsule, receiving_privkey,
-                            delegating_pubkey, signing_pubkey)
+    cleartext = pre.decrypt(ciphertext, capsule, receiving_privkey)
     assert some_dict[capsule] == "Thing that Bob wants to try per-Capsule"
     assert cleartext == plain_data
 
@@ -132,13 +131,13 @@ def test_capsule_length(alices_keys, bobs_keys):
 
     priv_key_bob, pub_key_bob = bobs_keys
 
-    sym_key, capsule = pre._encapsulate(delegating_privkey.get_pubkey().point_key)
+    sym_key, capsule = pre._encapsulate(delegating_privkey.get_pubkey())
+
+    kfrags = pre.split_rekey(delegating_privkey, signer, pub_key_bob, 10, 15)
 
     capsule.set_correctness_keys(delegating=delegating_privkey.get_pubkey(),
                                  receiving=pub_key_bob,
                                  verifying=signing_privkey.get_pubkey())
-
-    kfrags = pre.split_rekey(delegating_privkey, signer, pub_key_bob, 10, 15)
 
     for counter, kfrag in enumerate(kfrags):
         assert len(capsule) == counter
--- a/tests/test_capsule/test_capsule_operations.py
+++ b/tests/test_capsule/test_capsule_operations.py
@@ -6,14 +6,21 @@ from umbral.point import Point
 from umbral.pre import Capsule
 from umbral.signing import Signer
 from umbral.keys import UmbralPrivateKey
+from umbral.config import default_params
 
 
 def test_capsule_creation(alices_keys):
+
+    params = default_params()
+
     with pytest.raises(TypeError):
-        rare_capsule = Capsule()  # Alice cannot make a capsule this way.
+        rare_capsule = Capsule(params)  # Alice cannot make a capsule this way.
+
+
 
     # Some users may create capsules their own way.
-    custom_capsule = Capsule(point_e=Point.gen_rand(),
+    custom_capsule = Capsule(params,
+                             point_e=Point.gen_rand(),
                              point_v=Point.gen_rand(),
                              bn_sig=CurveBN.gen_rand())
 
@@ -28,17 +35,22 @@ def test_capsule_creation(alices_keys):
 
 
 def test_capsule_equality():
-    one_capsule = Capsule(point_e=Point.gen_rand(),
+    params = default_params()
+
+    one_capsule = Capsule(params,
+                          point_e=Point.gen_rand(),
                           point_v=Point.gen_rand(),
                           bn_sig=CurveBN.gen_rand())
 
-    another_capsule = Capsule(point_e=Point.gen_rand(),
+    another_capsule = Capsule(params,
+                              point_e=Point.gen_rand(),
                               point_v=Point.gen_rand(),
                               bn_sig=CurveBN.gen_rand())
 
     assert one_capsule != another_capsule
 
-    activated_capsule = Capsule(point_e_prime=Point.gen_rand(),
+    activated_capsule = Capsule(params,
+                                point_e_prime=Point.gen_rand(),
                                 point_v_prime=Point.gen_rand(),
                                 point_noninteractive=Point.gen_rand())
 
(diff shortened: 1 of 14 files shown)


umbral/pre.py
```diff
@@ -108,15 +111,17 @@ def from_bytes(cls, capsule_bytes: bytes, curve: ec.EllipticCurve = None):
...
                 (Point, point_size),  # point_e
                 (Point, point_size),  # point_v
                 (CurveBN, bn_size)  # bn_sig
             )
+        else:
+            raise ValueError("Byte string does not have a valid length for a Capsule")
 
         components = splitter(capsule_bytes)
-        return cls(*components)
+        return cls(*components, params=params)
```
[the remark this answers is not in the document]
cygnusv: Yep, that's why I'm leaning towards removing redundant `params` as argument in some places (assuming there's a capsule or other structure already holding `params`) and making it non-optional in others (e.g., in Umbral key generation).

tests/test_capsule/test_capsule_operations.py
```diff
@@ -28,43 +35,52 @@ def test_capsule_creation(alices_keys):
 
 
 def test_capsule_equality():
-    one_capsule = Capsule(point_e=Point.gen_rand(),
+    params = default_params()
```
tuxxy: We pass this in a ton. Should we make this a pytest fixture?

tests/test_capsule/test_capsule_serializers.py
```diff
@@ -39,26 +40,29 @@ def test_activated_capsule_serialization(alices_keys, bobs_keys):
     delegating_pubkey = delegating_privkey.get_pubkey()
     signer_alice = Signer(signing_privkey)
 
-    receiving_privkey, receiving_pubkey = bobs_keys
+    params = delegating_privkey.params
```
tuxxy: This is an interesting paradigm where each key and capsule can have its own params.

This might be a great pattern in the future, but we should be careful with it for now.

tests/test_simple_api.py
```diff
@@ -15,54 +17,160 @@
...
 
 
 @pytest.mark.parametrize("curve", secp_curves)
 @pytest.mark.parametrize("N, M", parameters)
 def test_simple_api_on_multiple_curves(N, M, curve):
+    test_simple_api(N, M, curve)
+
+
+@pytest.mark.parametrize("N, M", parameters)
+def test_lifecycle_with_serialization(N, M, curve=default_curve()):
```
tuxxy: 👍 Looks really good.

umbral/fragments.py
```diff
@@ -148,7 +150,7 @@ def to_bytes(self) -> bytes:
 
         return result
 
-    def _bn_keytes__(self):
+    def _bytes__(self):
```
tuxxy: This needs to be `__bytes__` :)

umbral/keys.py
```diff
@@ -117,6 +114,12 @@ def to_bytes(self, password: bytes=None, _scrypt_cost: int=20,
 
         return umbral_privkey
 
+    def __bytes__(self):
```
tuxxy: We don't offer `__bytes__` on `UmbralPrivateKey` to prevent accidental misuse.

I think this should be removed.

umbral/keys.py
```diff
@@ -82,8 +79,8 @@ def from_bytes(cls, key_bytes: bytes, params: UmbralParameters=None,
         bn_key = CurveBN.from_bytes(key_bytes, params.curve)
         return cls(bn_key, params)
 
-    def to_bytes(self, password: bytes=None, _scrypt_cost: int=20,
-                 encoder: Callable=None):
+    def to_bytes(self, password: bytes = None, _scrypt_cost: int = 20,
```
tuxxy: In instances where you set a default value with type in the function like the `_scrypt_cost` variable here, the equal sign should have no spaces such that:

`_scrypt_cost: int=20`

In cases where there is no type, such as `None`, it should have a space:

`password: bytes = None`

https://www.python.org/dev/peps/pep-0008/#other-recommendations

I'm not necessarily a huge fan of this PEP8 rule, as such I will leave it up to everyone else to decide how to do this. Generally, I'm in favor of not doing this as it just isn't necessary.

umbral/signing.py
```diff
@@ -21,7 +20,7 @@ class Signature:
     between (r, s) and DER formatting.
     """
 
-    def __init__(self, r: int, s: int):
+    def __init__(self, r: CurveBN, s: CurveBN):
```
tuxxy: If we're going to use `CurveBN` here, this should be in a separate PR. I'm not quite sure what the best options here are yet and I want to think about it more.

umbral/signing.py
```diff
@@ -56,23 +55,27 @@ def verify(self, message: bytes, verifying_key: UmbralPublicKey) -> bool:
...
+        
+        return cls(CurveBN.from_int(r, curve), CurveBN.from_int(s, curve))
 
     def _der_encoded_bytes(self):
-        return encode_dss_signature(self.r, self.s)
+        return encode_dss_signature(int(self.r), int(self.s))
 
     def __bytes__(self):
-        return self.r.to_bytes(32, "big") + self.s.to_bytes(32, "big")
+        return self.r.to_bytes() + self.s.to_bytes()
```
tuxxy: 👍

umbral/keys.py
jMyles: I agree with @tuxxy here.  What's the use case?

umbral/keys.py
jMyles: I used to think this was a really dumb style, but then I came around to it.  The thing is: in kwargs, `x=y` looks like assignment.  With the space around it, your eyes learn to treat the space as a type hint instead of argument assignment.

umbral/signing.py
jMyles: Yep, agreed.

umbral/keys.py
tuxxy: 🤔 But it's not a type hint? It's a default assignment if the argument is not passed?

jMyles: Eh?  It is a type hint.  Without the space, it looks like you're assigning the value `None` to the name `bytes`.

tuxxy: Oh, I see! Gotcha.

tests/test_capsule/test_capsule_operations.py
jMyles: Yes.

tests/test_capsule/test_capsule_operations.py
```diff
@@ -28,43 +35,52 @@ def test_capsule_creation(alices_keys):
...
+
+    one_capsule = Capsule(params,
```
jMyles: Wow - big change.  

This is pretty cool I think.

But I'm left wondering: if `Capsule` is going to *require* params, does it perhaps make sense to make it private?  Is there really ever a public reason for someone to init a `Capsule`?

tests/test_capsule/test_capsule_operations.py
```diff
@@ -28,43 +35,52 @@ def test_capsule_creation(alices_keys):
...
     assert activated_capsule != one_capsule
 
 
 def test_decapsulation_by_alice(alices_keys):
+    params = default_params()
+
     delegating_privkey, _signing_privkey = alices_keys
 
-    sym_key, capsule = pre._encapsulate(delegating_privkey.get_pubkey().point_key)
+    sym_key, capsule = pre._encapsulate(delegating_privkey.get_pubkey())
```
jMyles: Oh hallelujah.  Holy shit what a relief.

tests/test_capsule/test_capsule_operations.py
```diff
@@ -28,43 +35,52 @@ def test_capsule_creation(alices_keys):
...
     assert len(sym_key) == 32
 
     # The symmetric key sym_key is perhaps used for block cipher here in a real-world scenario.
-    sym_key_2 = pre._decapsulate_original(delegating_privkey.bn_key, capsule)
+    sym_key_2 = pre._decapsulate_original(delegating_privkey, capsule)
```
jMyles: oh my goodness this is seriously amazing.  This is how pyUmbral has always wanted to read.  (Although again, as I outlined in #161, I might advocate even going all the way down to just `delegating_key` or `delegating_secret` and dropping the "priv" idea).

tests/test_capsule/test_capsule_serializers.py
jMyles: Agreed on both points 100%.  Interesting but be careful.

tests/test_capsule/test_capsule_operations.py
cygnusv: Yes, we should. Perhaps in another PR?

umbral/fragments.py
cygnusv: :heavy_check_mark:

umbral/keys.py
cygnusv: None. It was a silly move :man_facepalming: ... Fixed.

PR #162, opened 2018-06-05
Serialization fixes for multicurve

cygnusv:
## What this does ##

* Adds a new set of tests in `test_simple_api` that goes through the Umbral lifecycle with intermediate serialization/deserialization, in order to better capture a real use of pyUmbral.
* Several bugfixes: 
    * Fixes `Signature` to account for the order of the curve (hence adding support to non-default curves)
         * Uses `CurveBN` instead of `int`
         * `Signature.from_bytes` takes an additional argument 'curve'
    * Fixes a bug in the deserialization method in `Capsule`, `KFrag`, `CapsuleFrag`, and `CorrectnessProof` for non-default curves, due to not passing the curve to `ByteStringSplitter`.

Note: Depends on #159

### DIFF
--- a/umbral/_pre.py
+++ b/umbral/_pre.py
@@ -7,10 +7,9 @@ from umbral.params import UmbralParameters
 def prove_cfrag_correctness(cfrag: "CapsuleFrag",
                             kfrag: "KFrag",
                             capsule: "Capsule",
-                            metadata: bytes = None,
-                            params: UmbralParameters = None
+                            metadata: bytes = None
                             ) -> "CorrectnessProof":
-    params = params if params is not None else default_params()
+    params = capsule._umbral_params
 
     rk = kfrag._bn_key
     t = CurveBN.gen_rand(params.curve)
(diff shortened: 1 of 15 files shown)


umbral/_pre.py
```diff
@@ -7,10 +7,9 @@
 def prove_cfrag_correctness(cfrag: "CapsuleFrag",
                             kfrag: "KFrag",
                             capsule: "Capsule",
-                            metadata: bytes = None,
-                            params: UmbralParameters = None
+                            metadata: bytes = None
                             ) -> "CorrectnessProof":
-    params = params if params is not None else default_params()
+    params = capsule._umbral_params
```
jMyles: Yeah, see, maybe this is wanting to not be private.

umbral/params.py
```diff
@@ -21,3 +21,14 @@ def __init__(self, curve: ec.EllipticCurve):
 
         parameters_seed = b'NuCypherKMS/UmbralParameters/'
         self.u = unsafe_hash_to_point(g_bytes, self, parameters_seed + b'u')
+
+    def __eq__(self, other):
+
```
jMyles: Any concern about execution time here?

umbral/signing.py
```diff
@@ -56,23 +54,26 @@ def verify(self, message: bytes, verifying_key: UmbralPublicKey) -> bool:
...
+        
+        return cls(CurveBN.from_int(r, curve), CurveBN.from_int(s, curve))
 
     def _der_encoded_bytes(self):
-        return encode_dss_signature(self.r, self.s)
+        return encode_dss_signature(int(self.r), int(self.s))
 
     def __bytes__(self):
-        return self.r.to_bytes(32, "big") + self.s.to_bytes(32, "big")
+        return self.r.to_bytes() + self.s.to_bytes()
```
jMyles: Jeez that's nice.

PR #169, opened 2018-06-24
Implements __eq__ and __hash__ on KFrag, is_compressed in UmbralPublicKey.to_bytes()

jMyles:
* Implements `__eq__` and `__hash__` on `KFrag`
* Allows `is_compressed=False` in `UmbralPublicKey.to_bytes()`

### DIFF
--- a/umbral/fragments.py
+++ b/umbral/fragments.py
@@ -2,12 +2,10 @@ from bytestring_splitter import BytestringSplitter
 from cryptography.hazmat.primitives.asymmetric import ec
 
 from umbral._pre import assess_cfrag_correctness, verify_kfrag
-from umbral.config import default_curve, default_params
+from umbral.config import default_curve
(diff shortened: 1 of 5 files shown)


umbral/fragments.py
```diff
@@ -77,6 +77,12 @@ def verify(self,
     def __bytes__(self):
         return self.to_bytes()
 
+    def __eq__(self, other):
+        return bytes(self) == bytes(other)
```
cygnusv: This isn't constant-time, isn't it?

umbral/fragments.py
```diff
@@ -77,6 +77,12 @@ def verify(self,
...
+
+    def __hash__(self):
+        return hash(bytes(self))
+
```
cygnusv: Better just `hash(bytes(self._id))`

PR #177, opened 2018-06-28
CircleCI: Mypy Type Check + Pipenv Lock

KPrasch:
- Adds mypy build with HTML report generation
- Locks pip dependencies 
- Changes dependency caching key: Use only a single dependency bundle per workflow (for speed and accurate test layers)
- Create base python job templates


michwill:
Yeah, I'd keep the ability to run tests locally as long as you spinned up the dev environment

PR #178, opened 2018-07-01
Refactor pyUmbral to use Curve object

tuxxy:
### What this does:
1. Creates a `Curve` object that uses a whitelist as to which curves can be instantiated.
   1. SECP256K1
   2. SECP256R1 (P-256)
   3. SECP384R1 (P-384)
2. Adds `SECP256K1`, `SECP256R1`, and `SECP384R1` curve objects.
3. Refactors the `openssl` module to use `curve` objects instead of `EC_GROUP` structs.
4. The `params` and `config` modules return an already instantiated `Curve` object. This prevents unnecessary creations of an `EC_GROUP` struct.
5. Items 3 & 4 _significantly_ reduce the memory consumption of pyUmbral. The memory leak is _no more_.
6. When creating `CurveBN` and `Point` objects, you no longer pass in a `group`, `order`, and `curve_nid`. Instead, you pass in an already instantiated `Curve` object.
7. Multi-curve support is now significantly cleaner.

### DIFF
--- a/tests/test_curve.py
+++ b/tests/test_curve.py
@@ -1,28 +1,29 @@
 import pytest
-from umbral.curve import Curve, SECP256R1, SECP256K1, SECP384R1
+from umbral.curve import Curve, SECP256R1, SECP256K1, SECP384R1, _AVAIL_CURVES
 
 
 def test_curve_whitelist():
     # Test the AVAIL_CURVES dict to have only these three curves:
-    assert len(Curve._Curve__AVAIL_CURVES) == 3
-    assert Curve._Curve__AVAIL_CURVES['secp256r1'] == 415
-    assert Curve._Curve__AVAIL_CURVES['secp256k1'] == 714
-    assert Curve._Curve__AVAIL_CURVES['secp384r1'] == 715
+    assert len(_AVAIL_CURVES) == 3
+    assert _AVAIL_CURVES['secp256r1'] == 415
+    assert _AVAIL_CURVES['secp256k1'] == 714
+    assert _AVAIL_CURVES['secp384r1'] == 715
 
     # Test that we can't instantiate other curves:
     with pytest.raises(ValueError):
         Curve(711)
 
     # Test the hardcoded curves are what they're supposed to be:
-    test_p256 = SECP256R1()
-    test_secp256k1 = SECP256K1()
-    test_p384 = SECP384R1()
-
-    # Test the supported curves property
-    assert test_p256.supported_curves\
-    == test_secp256k1.supported_curves\
-    == test_p384.supported_curves
+    test_p256 = SECP256R1
+    test_secp256k1 = SECP256K1
+    test_p384 = SECP384R1
 
     assert test_p256.curve_nid == 415
     assert test_secp256k1.curve_nid == 714
     assert test_p384.curve_nid == 715
+
+    # Test the supported curves property
+    assert test_p256.supported_curves == _AVAIL_CURVES
+    assert test_secp256k1.supported_curves == _AVAIL_CURVES
+    assert test_p384.supported_curves == _AVAIL_CURVES
+
--- a/tests/test_curve.py
+++ b/tests/test_curve.py
@@ -1,5 +1,4 @@
 import pytest
-
 from umbral.curve import Curve, SECP256R1, SECP256K1, SECP384R1
 
 
--- a/tests/test_curve.py
+++ b/tests/test_curve.py
@@ -1,0 +1,24 @@
+import pytest
+
+from umbral.curve import Curve, SECP256R1, SECP256K1, SECP384R1
+
+
+def test_curve_whitelist():
+    # Test the AVAIL_CURVES dict to have only these three curves:
+    assert len(Curve.AVAIL_CURVES) == 3
+    assert Curve.AVAIL_CURVES['secp256r1'] == 415
+    assert Curve.AVAIL_CURVES['secp256k1'] == 714
+    assert Curve.AVAIL_CURVES['secp384r1'] == 715
+
+    # Test that we can't instantiate other curves:
+    with pytest.raises(ValueError):
+        Curve(711)
+
+    # Test the hardcoded curves are what they're supposed to be:
+    test_p256 = SECP256R1()
+    test_secp256k1 = SECP256K1()
+    test_p384 = SECP384R1()
+
+    assert test_p256.curve_nid == 415
+    assert test_secp256k1.curve_nid == 714
+    assert test_p384.curve_nid == 715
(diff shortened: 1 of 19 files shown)


tests/test_curve.py
```diff
@@ -0,0 +1,28 @@
...
+    # Test that we can't instantiate other curves:
+    with pytest.raises(ValueError):
+        Curve(711)
+
+    # Test the hardcoded curves are what they're supposed to be:
+    test_p256 = SECP256R1()
+    test_secp256k1 = SECP256K1()
+    test_p384 = SECP384R1()
+
+    # Test the supported curves property
```
jMyles: Eh?  What are we showing here?

tuxxy: They're all caps because they act as constants. Nothing in the structure actually changes, there are no modifying methods.

If by importing the instances you mean creating a new instance with every import, I would disagree. This `Curve` object contains the `EC_GROUP` which is the primary cause of memory issues in pyUmbral. Instantiating it over and over again is the result of the memory leak just fixed.

Currently, we access the curve through the `Config` interface via the `default_curve` or `default_params` functions. This API still preserves this architecture, but I don't like it. I'm thinking that there should be some object that contains the params, the config, and the curve all in one.

cygnusv:
Just for the record: 

**Before:**
screenshot from 2018-05-30 15-19-07
**After:**
screenshot from 2018-07-02 13-23-42

jarl

umbral/openssl.py
```diff
@@ -20,67 +22,62 @@ def _get_new_BN(set_consttime_flag=True):
 
 def _get_ec_group_by_curve_nid(curve_nid: int):
     """
-    Returns the group of a given curve via its OpenSSL nid.
+    Returns the group of a given curve via its OpenSSL nid. This must be freed
+    after each use otherwise it leaks memory.
     """
     group = backend._lib.EC_GROUP_new_by_curve_name(curve_nid)
     backend.openssl_assert(group != backend._ffi.NULL)
```
cygnusv: Isn't here missing an `EC_GROUP_clear_free`?

umbral/point.py
```diff
@@ -124,10 +119,7 @@ def to_bytes(self, is_compressed=True):
         if is_compressed is set to True.
         """
         affine_x, affine_y = self.to_affine()
-
-        # TODO: Get size of curve via OpenSSL
-        order = openssl._get_ec_order_by_curve_nid(self.curve.curve_nid)
-        key_size = backend._lib.BN_num_bytes(order)
+        key_size = get_field_order_size_in_bytes(self.curve)
```
cygnusv: Good catch, man! I think it's better to call that `coord_size` or `field_size`, as in the `from_bytes()` method.

umbral/curve.py
```diff
@@ -0,0 +1,63 @@
...
+        'secp384r1': 715,
+    }
+
+    def __init__(self, curve_nid: int):
+        """
+        Instantiates an OpenSSL curve with the provided curve_nid and derives
+        the proper EC_GROUP struct and order. You can _only_ instantiate curves
+        with supported nids (see `Curve.supported_curves`).
+        """
+        if curve_nid not in self.__AVAIL_CURVES.values():
```
KPrasch: Looks like this is intentionally used by NID only?

umbral/curve.py
```diff
@@ -0,0 +1,63 @@
+from umbral import openssl
+
+
+class Curve:
```
KPrasch: No EC?

KPrasch:
This memory leak fix is critical - Im also working on a better run environment for valgrind using the procedure found here: https://pythonextensionpatterns.readthedocs.io/en/latest/debugging/valgrind.html#building-python-for-valgrind.

umbral/openssl.py
tuxxy: If you call it, it gets cleared from memory immediately. Apparently, the garbage collector picks it up right then.

I think I would add it to a `__del__` method, but not really sure what the benefits are.

umbral/curve.py
```diff
@@ -0,0 +1,63 @@
...
+    as a convenient whitelist to limit the curves used in pyUmbral.
+    """
+
+    __AVAIL_CURVES = {
+        'secp256r1': 415,
+        'secp256k1': 714,
+        'secp384r1': 715,
+    }
+
+    def __init__(self, curve_nid: int):
```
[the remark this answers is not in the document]
tuxxy: Using the subclasses feels much more cleaner. I think this is apparent in the tests. It's also pretty apparent which curves are going to be supported for a while.

Your second concern is valid, but for it to be a problem, a curve will have to be instantiated many many times. To cause a leak as bad as the previous one, then the user will have to instantiate each curve as much as we created an `EC_GROUP` in the previous version.

We could fix this by adding a class that stores instances of curves, but I suspect this is adding complexity for very little return. Documentation for this must simply state that instantiating multiple curves unnecessarily will cause memory issues.

Another idea is that we implement `__del__` that clears the used memory.

umbral/curve.py
tuxxy: Yep, it's the method to creating an `EC_GROUP` struct. All the other calls require the `EC_GROUP` and this is the only place to get it. I'm evaluating whether or not we even need to store the `curve_nid` to prevent misuse.

umbral/curve.py
tuxxy: How do you mean?

umbral/curve.py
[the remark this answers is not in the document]
jMyles: > Using the subclasses feels much more cleaner. I think this is apparent in the tests

No it doesn't and no it isn't.  These things aren't classes.  This is clearly not the correct way to do this.  And even if it were, overriding __init__ is not the correct way to do what you're trying to do - if it were, you'd use class attributes.  

> We could fix this by adding a class that stores instances of curves

Wait, you mean a registry?  For what?  You already have a module that can do that.

There's no need for any of this complexity.  init three `Curve` instances and then people will import them.  Done.  Simple and consistent.

tests/test_curve.py
jMyles: > They're all caps because they act as constants.

But... they don't.  The user can (intentionally or otherwise) init a bunch of them.  They're not constants - and why not?  Let's make them constants. 

> If by importing the instances you mean creating a new instance with every import,

Wait, what?  I didn't say anything of the sort.  I don't even know how you'd do that - with some importlib gymnastics or something?  Anyway, no I'm saying just make them instances instead of classes.

KPrasch:
Agreed - What about renaming `Curve` -> `__Curve` - then instantiating three curve instances with all caps names for others to import.  Module-level singularity is sufficient here.

tuxxy:
I don't agree with making `Curve` unnecessarily private (which it becomes, in this case, imho.)

I will go ahead and make them singletons, though.

KPrasch:
Making these singeltons may lead to a get or create style pattern, sort of lazy.

KPrasch:
If you want to run tests against python 3.7- you can rebase over #188

Issue #183, opened 2018-07-02
mypy: Staticly Typed pyUmbral


KPrasch:
Currently automated mypy builds result in numerous warnings, which we aim to reduce to zero.  
https://249-115219228-gh.circle-artifacts.com/0/home/circleci/pyUmbral-36/mypy_report/index.html

Some initial thoughts:
- We can control how strict mypy check are: https://mypy.readthedocs.io/en/latest/config_file.html
- We will need stubs for Ctypes: https://www.python.org/dev/peps/pep-0484/#stub-files
- We may be able to use MonkeyType or another similar tool to assist: https://github.com/Instagram/MonkeyType

PR #189, opened 2018-07-06
PEP 484 Type Hints for pyUmbral

KPrasch:
#### Goal
Passing mypy checks, PEP 484 compliance.

##### Adds type annotations to the following modules:
- [X] `_pre.py`
- [X] `config.py`
- [ ] `curvebn.py`
- [X] `dem.py`
- [x] `fragments.py`
- [X] `keys.py`
- [x] `opensll.py` - Skipped
- [X] `params.py`
- [ ] `point.py`
- [X] `pre.py`
- [X] `signing.py`
- [X] `utils.py`

See: #183

### DIFF
--- a/umbral/_pre.py
+++ b/umbral/_pre.py
@@ -1,20 +1,21 @@
+from typing import Optional
+
 from umbral.curvebn import CurveBN
-from umbral.config import default_params
 from umbral.keys import UmbralPublicKey
-from umbral.params import UmbralParameters
 
 
-def prove_cfrag_correctness(cfrag: "CapsuleFrag",
-                            kfrag: "KFrag",
-                            capsule: "Capsule",
-                            metadata: bytes = None
-                            ) -> "CorrectnessProof":
+def prove_cfrag_correctness(cfrag: 'CapsuleFrag',
+                            kfrag: 'KFrag',
+                            capsule: 'Capsule',
+                            metadata: Optional[bytes] = None
+                            ) -> 'CorrectnessProof':
+
     params = capsule._umbral_params
 
     rk = kfrag._bn_key
     t = CurveBN.gen_rand(params.curve)
     ####
-    ## Here are the formulaic constituents shared with `assess_cfrag_correctness`.
+    # Here are the formulaic constituents shared with `assess_cfrag_correctness`.
     ####
     e = capsule._point_e
     v = capsule._point_v
@@ -45,7 +46,7 @@ def prove_cfrag_correctness(cfrag: "CapsuleFrag",
         raise capsule.NotValid("Capsule verification failed.")
 
 
-def assess_cfrag_correctness(cfrag, capsule: "Capsule"):
+def assess_cfrag_correctness(cfrag: 'CapsuleFrag', capsule: 'Capsule') -> bool:
 
     correctness_keys = capsule.get_correctness_keys()
 
@@ -62,7 +63,7 @@ def assess_cfrag_correctness(cfrag, capsule: "Capsule"):
     params = capsule._umbral_params
 
     ####
-    ## Here are the formulaic constituents shared with `prove_cfrag_correctness`.
+    # Here are the formulaic constituents shared with `prove_cfrag_correctness`.
     ####
     e = capsule._point_e
     v = capsule._point_v
@@ -110,11 +111,11 @@ def assess_cfrag_correctness(cfrag, capsule: "Capsule"):
            & correct_rk_commitment
 
 
-def verify_kfrag(kfrag,
+def verify_kfrag(kfrag: 'KFrag',
                  delegating_pubkey: UmbralPublicKey,
-                 signing_pubkey,
+                 signing_pubkey: UmbralPublicKey,
                  receiving_pubkey: UmbralPublicKey
-                 ):
+                 ) -> bool:
 
 
     params = delegating_pubkey.params
--- a/umbral/config.py
+++ b/umbral/config.py
@@ -1,6 +1,8 @@
+from typing import Optional, Type, Union
 from warnings import warn
 
 from umbral.curve import Curve, SECP256K1
+from umbral.params import UmbralParameters
 
 
 class _CONFIG:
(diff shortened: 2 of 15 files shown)


KPrasch:
The remaining modules (curvebn, openssl, and point) use ctypes that are a little harder to annotate.  Perhaps we can introduce types with > 1 Pull Request.

tuxxy:
> The remaining modules (curvebn, openssl, and point) use ctypes that are a little harder to annotate.

Yes, I think it's just `CData`. Given the generic name, I have tried to name the params to be similar to the C structs that are passed in. IE: `EC_GROUP`.

KPrasch:
We dont have 100% coverage or accuracy - but mypy errors have been vastly reduced.  

I'd like to take this PR off WIP while we work on annotating the remaining parts of the codebase in a future PR.

umbral/pre.py
```diff
@@ -1,45 +1,47 @@
...
-                 point_noninteractive=None,
-                 delegating_pubkey: UmbralPublicKey = None,
-                 receiving_pubkey: UmbralPublicKey = None,
-                 verifying_pubkey=None):
+                 point_e: Optional[Point] = None,
+                 point_v: Optional[Union[int, Point]] = None,
+                 bn_sig: Optional[Union[int, CurveBN]] = None,
+                 point_e_prime: Optional[Point] = None,
+                 point_v_prime: Optional[Union[int, Point]] = None,
+                 point_noninteractive: Optional[Union[int, Point]] = None,
```
cygnusv: All `Points` above are just `Optional[Point]`, not `Optional[Union[int, Point]]`. Same for `bn_sig`, but with `CurveBN`.

umbral/pre.py
```diff
@@ -177,7 +180,8 @@ def verify(self) -> bool:
         s = self._bn_sig
         h = CurveBN.hash(e, v, params=self._umbral_params)
 
-        return s * g == v + (h * e)
+        result = s * g == v + (h * e)      # type: bool
```
cygnusv: Isn't this redundant? We already said above that the result is `bool`.

umbral/pre.py
```diff
@@ -260,15 +264,16 @@ def __eq__(self, other):
             # Again, it's hard to imagine why.
             return False
 
-    def __hash__(self):
+    @typing.no_type_check
```
cygnusv: Just curious, why?

umbral/signing.py
```diff
@@ -81,22 +82,22 @@ def __len__(self):
     def __add__(self, other):
```
cygnusv: Why `__add__` and `__len__` don't have type hints, but `__bytes__`, `__radd__', and `__eq__` do?

umbral/utils.py
```diff
@@ -1,12 +1,16 @@
...
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric.ec import SECP192R1, SECP256K1, SECP256R1, SECP384R1
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
-from cryptography.hazmat.primitives.asymmetric import ec
+
 from umbral import openssl
 
-def lambda_coeff(id_i, selected_ids):
+CurveTypes = Union[Type[SECP256K1], Type[SECP192R1], Type[SECP384R1], Type[SECP256R1]]
```
cygnusv: We can get rid of `SECP192R1`.

KPrasch:
Cool - Ill peck away at the changes.

 Feel free to add more comments  - I combed through the changeset - but much of the annotations were auto-generated with monkeytype so more eyes on the code is a good thing.

umbral/signing.py
KPrasch: other than `__eq__`, They just haven't been covered yet.  As for `__eq__` an implementation change may be needed as `__eq__` needs to be able to take any `object`.

umbral/utils.py
KPrasch: Well, it's a type that is passed in during test runtime, an it needs to be included in the annotation for it to pass type checks.  Perhaps we need curves type-aliases

umbral/pre.py
KPrasch: there is a tricky check inside the genexp here.

umbral/pre.py
KPrasch: Yeah - perhaps redundant - Looking for precision on the charts, the return value type and the return type of the arithmetic operators are different.

umbral/fragments.py
```diff
@@ -69,19 +69,19 @@ def to_bytes(self):
         return self._id + key + ni + commitment + xcoord + signature
 
     def verify(self,
-               signing_pubkey,
+               signing_pubkey: UmbralPublicKey,
```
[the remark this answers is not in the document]
KPrasch: Perhaps a type alias, skip,  or `Any` will suffice

umbral/config.py
```diff
@@ -18,19 +22,19 @@ def __set_curve_by_default(cls):
...
     @classmethod
-    def params(cls):
+    def params(cls) -> UmbralParameters:
         if not cls.__params:
             cls.__set_curve_by_default()
         return cls.__params
 
     @classmethod
-    def curve(cls):
+    def curve(cls) -> Union[Type[SECP256R1], Type[SECP256K1]]:
```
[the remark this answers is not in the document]
KPrasch: Yeah - Sounds like a good Idea - we need a single sourced alias for curves, IMO

umbral/_pre.py
```diff
@@ -110,11 +111,11 @@ def assess_cfrag_correctness(cfrag, capsule: "Capsule"):
            & correct_rk_commitment
 
 
-def verify_kfrag(kfrag,
+def verify_kfrag(kfrag: 'KFrag',
                  delegating_pubkey: UmbralPublicKey,
-                 signing_pubkey,
+                 signing_pubkey: UmbralPublicKey,
```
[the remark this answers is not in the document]
KPrasch: Hmm, what shall we do in the meantime?

KPrasch:
What do you all think about addressing the request for changes in a future PR?

umbral/pre.py
KPrasch: Yeah - @jMyles and were are a bit stumped on how int is passed here during test-runtime.

umbral/_pre.py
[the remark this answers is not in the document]
KPrasch: Yep - looks like we need to implement a custom subprtocol: https://mypy.readthedocs.io/en/latest/protocols.html#simple-user-defined-protocols

umbral/pre.py
cygnusv: Different? I don't see why they're different.

PR #190, opened 2018-07-08
[WIP]Pre-version Housekeeping

KPrasch:
- Establishes bumpversion configuration 
- Single-sources project version and metadata


setup.py
```diff
@@ -1,19 +1,58 @@
...
-      extras_require={'testing': TESTS_REQUIRE},
+    'pytest-mypy',
+    'pytest-mock',
+    'mock',
+    'coverage',
+    'codecov',
+    'monkeytype',
+    ]
+
+INSTALL_REQUIRES = ['msgpack-python',
```
michwill: Wait. Why is this list not the same as in `Pipfile`?

KPrasch: Hmm - This is certainly not correct. good catch.

setup.py
```diff
@@ -1,19 +1,58 @@
...
 TESTS_REQUIRE = [
     'pytest',
-    'coverage',
     'pytest-cov',
-    'pdbpp',
-    'ipython'
-]
-
-setup(name='umbral',
-      version='0.1',
```
cygnusv: All the time, it was...`version='0.1'`..... YOU MANIACS! YOU BLEW IT UP!

KPrasch:
Just waiting on the initial `bytestringSplittler` release to untag this one.

setup.py
KPrasch: Heh, we switched over to bumpversion

PR #191, opened 2018-07-09
PEP 484 mypy type checks (Clone of #189, but rebased over master w/ Curve API)

tuxxy:
https://github.com/nucypher/pyUmbral/pull/189


tuxxy:
This PR will not be used.

PR #192, opened 2018-07-11
Minor improvements to CurveBN

cygnusv:
## What this does ##
- Introduces `CurveBN.__neg__`, the modular opposite unary operation. 
- Moves the function `get_field_order_size_in_bytes` from `umbral.utils` to a method of `Curve`. This makes things cleaner in CurveBN and Point, as well as removes forward references to CurveBN in type annotations of `umbral.utils`.
- Adds support to `int` arguments in `Curve.__add__` and `Curve.__sub__`.
- Introduces some type annotations for `CurveBN`. 
- Fixed memory leak in `CurveBN.__truediv__`. It was a small leak (~ 5MB per 100K operations), not critical, but it was there. Nice before & after photos:
screenshot from 2018-07-11 11-28-09
screenshot from 2018-07-11 11-28-33

### DIFF
--- a/tests/test_primitives/test_bignum/test_bignum_arithmetic.py
+++ b/tests/test_primitives/test_bignum/test_bignum_arithmetic.py
@@ -1,3 +1,4 @@
+from cryptography.hazmat.backends.openssl import backend
 from umbral.curvebn import CurveBN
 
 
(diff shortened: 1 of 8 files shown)


tuxxy:
Not sure if I really like the idea of operations being performed on `int`s. I was thinking about pulling that out wherever it can be done.

Also, `__bytes__` hasn't been implemented because I like forcing the user to be very explicit in serializing the `CurveBN` to a bytestring.

michwill:
I think, I'd prefer `__bytes__` to be in though

tuxxy:
My concern is two primary things:
1. The user doesn't realize the bytestring they're using is the full `CurveBN`.
2. If any package ever automatically calls `__bytes__` for logging or _anything_, then it will leak data if it is private.

michwill:
The second is a very valid concern. Does logging ever call `__bytes__` or it uses `__str__ / __repr__`?

tuxxy:
All the time. It can use variations of those three (and perhaps others), but we ultimately don't control what appears in logs; therefore, I'd like to keep it simply as `CurveBN.to_bytes`.

cygnusv:
@tuxxy Reason 2 is enough for me to undo that commit. But I don't understand reason 1. Can you elaborate?

tuxxy:
I've brought this up many times, but I really don't like how python handles it's `bytes` API. You never know what is going to happen and it all depends on what you pass into the call to `bytes(..)`.

If you pass an `int` you get the `int` number of bytes of zero, by default. I think Python allowing overloading of this is dangerous and leads to readability issues.

PR #194, opened 2018-07-12
Some commits from 169 that slipped under the radar.

jMyles:
When #169 was merged into nucypher-depend, a few commits went by without ever being merged to master.  This PR rebases them on top of today's master.

Note: I have already force-pushed to nucypher-depend; if you are using that branch for development of nucypher, you'll need to forcefully integrate.

### DIFF
--- a/tests/test_keys/test_key_fragments.py
+++ b/tests/test_keys/test_key_fragments.py
@@ -25,6 +25,22 @@ def test_kfrag_serialization(alices_keys, bobs_keys):
     assert new_frag._point_xcoord == kfrags[0]._point_xcoord
 
 
+def test_kfrag_as_dict_key(alices_keys, bobs_keys):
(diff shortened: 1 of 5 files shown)


cygnusv:
What's the use case for `UmbralPublicKey.to_bytes(is_compressed=True)'? I'm asking this because in general we don't want to use uncompressed points anywhere.

jMyles:
The use case is as follows:

We save references to all nodes by address.  Since only decentralized characters have wallet addresses, we create addresses for federated characters by applying the same transformation normally applied to the ethereum keypair, but to the signing keypair instead.  This process requires uncompressed key material.

If you have a better idea about how to do this (that is, to track nodes with consistent logic even if they don't have an ethereum address), I'm certainly open to other ways.

cygnusv:
That makes total sense :+1:

PR #195, opened 2018-07-13
Umbral Test Vectors (I)

cygnusv:
## What this does ##

This PR defines a script (located in `vectors/generate_test_vectors.py`) that creates test vectors for some of the most basic primitives and cryptographic elements of pyUmbral:
- [X] CurveBN
- [X] Point
- [X] KFrags
- [X] CFrags

Test vectors are written as JSON files, and an initial version of these is located in the `vectors`folder. Note that it is currently impossible to generate deterministically some of these test vectors, so they will be randomized on each execution of the script; hence, we shouldn't automate the build of the test vectors. 

This PR also include tests to validate the test vectors. This can catch potential changes in pyUmbral that affect interoperability while still making the tests pass (e.g., changing some public parameter such as `params.u`).

A later PR will provide test vectors for:
- UmbralPublicKeys and UmbralPrivateKeys
- Capsules
- Ciphertexts
- Decryption of original capsules/ciphertexts
- Decryption of re-encrypted capsules

Other TODOs:
- Organize the script better (perhaps a CLI?)
- Investigate ways to generate the test vectors deterministically without being too invasive with the codebase (I want to avoid having to modify the internals of pyUmbral just for this)


tests/test_vectors.py
```diff
@@ -0,0 +1,170 @@
...
+    try:
+        with open(vector_file) as f:
+            vector_suite = json.load(f)
+    except OSError:
+        raise 
+
+    params = default_params()
+
+    for vector in vector_suite['vectors']:
+        hash_input = [bytes.fromhex(item['bytes']) for item in vector['input']]
```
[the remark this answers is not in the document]
michwill: Or do we just feed `bytes` here instead of actual classes to test hashing?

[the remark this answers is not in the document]
cygnusv: My initial intention was to provide a way to test how this hash function works, where the input can be bytes or other crypto objects. To this respect, @karce expressed similar concerns to yours in the associated goUmbral PR, since it would make the test to be more than a unit test, in the sense that it would depend on the correctness of serialization and deserialization of the criptographic elements. Since in the end the hash function transforms everything to bytes, tests can simply input the bytes representation and forget about the specified class, which can be seen just as auxiliary metadata. You can see an example of this working in the tests contained in this PR, which check the generated test vectors.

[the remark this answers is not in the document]
Karce: In goUmbral's implementation of these tests it simply ignores the class field since these classes aren't in goUmbral's implementation of Hash and it won't affect the result.

tests/test_vectors.py
```diff
@@ -0,0 +1,170 @@
...
+from umbral import pre
+
+def test_curvebn_operations():
+
+    vector_file = os.path.join('vectors', 'vectors_curvebn_operations.json')
+    try:
+        with open(vector_file) as f:
+            vector_suite = json.load(f)
+    except OSError:
+        raise
```
KPrasch: Is this a stub for future error handling?

vectors/generate_test_vectors.py
```diff
@@ -0,0 +1,272 @@
...
+        return bytes(data).hex()
+
+def create_test_vector_file(vector, filename, generate_again=False):
+    path = os.path.join(os.path.dirname(__file__), filename)
+
+    mode = 'w' if generate_again else 'x'
+    try:
+        with open(path, mode) as f:
+            json.dump(vector, f, indent=2)
+    except FileExistsError:
```
KPrasch: Hmm - when is this getting raised?

tests/test_vectors.py
cygnusv: Yeah, not sure how we should handle the fact that the test vector fail doesn't exist. Perhaps simply fail the test?

vectors/generate_test_vectors.py
cygnusv: The whole idea of this method is the following: Since the generation of test vectors is always randomized, new test vector files would be created every time the script runs, overwriting the old ones. I don't want that, since it could lead to the test not detecting a change between versions (e.g., changing the public parameters). So, we use the mode 'w' when we want to overwrite the file no matter what, and the mode 'x' when we only want to write the file if it doesn't exist. The latter raises FileExistsError if the file exist, which for us simply means that there's nothing to do here.

PR #196, opened 2018-07-16
Fixed incorrect Point operations 

cygnusv:
## What this does ##
- `Point` was only allowing deserialization of EC points whose affine coordinates where lower than the order of the curve, which is incorrect. Instead, coordinates must be lower than the order of the base field. This was affecting potentially more than 10<sup>38</sup> EC points, but the probability of hitting one of them by chance is in the order of 10<sup>-39</sup>, so this went unnoticed. The problem was affecting `Point.from_affine` and `Point.from_bytes`. Now is fixed.
- In pyUmbral we use the additive notation for EC operations, but the additive opposite of an EC point was implemented with `Point.__invert__` (which maps to the unary operator `~`) instead of the more correct `Point.__neg__` (which maps to the unary operator `-`, usually representing the additive opposite).

### DIFF
--- a/tests/test_primitives/test_point/test_point_arithmetic.py
+++ b/tests/test_primitives/test_point/test_point_arithmetic.py
@@ -5,10 +5,10 @@ from umbral.point import Point
 def test_mocked_openssl_point_arithmetic(mock_openssl, random_ec_point1, random_ec_point2, random_ec_curvebn1):
 
     operations_that_construct = (
-        random_ec_point1 * random_ec_curvebn1,  # __mul__
+        random_ec_point1 * random_ec_curvebn1, # __mul__
         random_ec_point1 + random_ec_point2,   # __add__
         random_ec_point1 - random_ec_point2,   # __sub__
-        -random_ec_point1                      # __invert__
+        -random_ec_point1                      # __neg__
     )
 
     with mock_openssl():
--- a/tests/test_primitives/test_point/test_point_arithmetic.py
+++ b/tests/test_primitives/test_point/test_point_arithmetic.py
@@ -8,7 +8,7 @@ def test_mocked_openssl_point_arithmetic(mock_openssl, random_ec_point1, random_
         random_ec_point1 * random_ec_curvebn1,  # __mul__
         random_ec_point1 + random_ec_point2,   # __add__
         random_ec_point1 - random_ec_point2,   # __sub__
-        ~random_ec_point1                      # __invert__
+        -random_ec_point1                      # __invert__
     )
 
     with mock_openssl():
--- a/tests/test_primitives/test_point/test_point_serializers.py
+++ b/tests/test_primitives/test_point/test_point_serializers.py
@@ -131,3 +131,18 @@ def test_serialize_point_at_infinity():
     # We want to catch specific InternalExceptions:
     # - Point at infinity (code 107)
     assert e.value.err_code[0].reason == 106
+
+
+def test_some_interesting_points():
+
+    compressed = 0x02fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2c
+    compressed = compressed.to_bytes(32+1, byteorder='big')
+
+    last_point = Point.from_bytes(compressed)
+
+    coords = (115792089237316195423570985008687907853269984665640564039457584007908834671660, 
+        109188863561374057667848968960504138135859662956057034999983532397866404169138)
+
+    assert last_point == Point.from_affine(coords)
+
+    # TODO: add point with x == 0 or y== 0
--- a/tests/test_primitives/test_point/test_point_serializers.py
+++ b/tests/test_primitives/test_point/test_point_serializers.py
@@ -129,20 +129,22 @@ def test_serialize_point_at_infinity():
         _bytes_point_at_infinity = point_at_infinity.to_bytes()
 
     # We want to catch specific InternalExceptions:
-    # - Point at infinity (code 107)
+    # - Point at infinity (code 106)
     assert e.value.err_code[0].reason == 106
 
 
-def test_some_interesting_points():
+def test_coords_with_special_characteristics():
 
+    # Testing that a point with the x coordinate greater than the curve order is still valid
     compressed = 0x02fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2c
     compressed = compressed.to_bytes(32+1, byteorder='big')
 
     last_point = Point.from_bytes(compressed)
 
+    # The same point, but obtained through the from_affine method
     coords = (115792089237316195423570985008687907853269984665640564039457584007908834671660, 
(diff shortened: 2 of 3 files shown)


tests/test_primitives/test_point/test_point_serializers.py
```diff
@@ -118,3 +118,31 @@ def test_point_not_on_curve():
...
+    
+    with pytest.raises(InternalError) as e:
+        _bytes_point_at_infinity = point_at_infinity.to_bytes()
+
+    # We want to catch specific InternalExceptions:
+    # - Point at infinity (code 107)
+    assert e.value.err_code[0].reason == 106
+
+
+def test_some_interesting_points():
```
tuxxy: Could probably use a more informative name, but I understand what's going on.

Maybe add some comments on _why_ those points are interesting?

tests/test_primitives/test_point/test_point_serializers.py
```diff
@@ -118,3 +118,31 @@ def test_point_not_on_curve():
...
+
+    p = Point.gen_rand()
+    point_at_infinity = p - p
+    
+    with pytest.raises(InternalError) as e:
+        _bytes_point_at_infinity = point_at_infinity.to_bytes()
+
+    # We want to catch specific InternalExceptions:
+    # - Point at infinity (code 107)
+    assert e.value.err_code[0].reason == 106
```
tuxxy: Do you think we should start catching these errors and build our own error system to raise specific errors from OpenSSL except cleaner?

Probably works better with building an arithmetic library.

umbral/point.py
```diff
@@ -194,11 +196,12 @@ def __sub__(self, other):
         """
         Performs subtraction by adding the inverse of the `other` to the point.
         """
-        return (self + (~other))
+        return (self + (-other))
 
-    def __invert__(self) -> 'Point':
+    def __neg__(self) -> 'Point':
```
tuxxy: I like this, but my only concern for this is that someone accidentally subs when they want to invert.

For example, you must do `(-foobar)` in this case to make it visually clear that you are doing an inversion.

Look at this line for example:
`a = b - -c`

This is a valid line, but it looks _weird._ The obvious answer is to do:
`a = b - (-c)`

It no longer looks _weird_, but I think it could be accidentally looked over. I like `__invert__` here because, while not mathematically correct, it looks far more interesting visually and I think would reduce the chances of a typo.

What do you think?

tests/test_primitives/test_point/test_point_arithmetic.py
```diff
@@ -8,7 +8,7 @@ def test_mocked_openssl_point_arithmetic(mock_openssl, random_ec_point1, random_
         random_ec_point1 * random_ec_curvebn1,  # __mul__
         random_ec_point1 + random_ec_point2,   # __add__
         random_ec_point1 - random_ec_point2,   # __sub__
-        ~random_ec_point1                      # __invert__
+        -random_ec_point1                      # __invert__
```
tuxxy: Missed the comment here.

umbral/point.py
```diff
@@ -66,10 +66,10 @@ def from_affine(cls, coords: Tuple[int, int], curve: Optional[Curve] = None) ->
 
         affine_x, affine_y = coords
         if type(affine_x) == int:
-            affine_x = openssl._int_to_bn(affine_x, curve=curve)
+            affine_x = openssl._int_to_bn(affine_x, curve=None)
 
         if type(affine_y) == int:
-            affine_y = openssl._int_to_bn(affine_y, curve=curve)
+            affine_y = openssl._int_to_bn(affine_y, curve=None)
```
tuxxy: :+1: Great catch

umbral/point.py
tuxxy: Also, if you do `a = b-c`, though not proper, is still valid and makes readability more problematic.

cygnusv: I think that choice would be far weirder. Python itself allows for expressions like `a = b - -c`

tuxxy: Yeah, but just because python allows doesn't make it good. If we're going down that road, there's a lot of horrible naughty things that Python allows.

cygnusv: We don't have to go down that road, true. But going back to the question, I really don't see what's the issue, or what do you want to protect users against. Using the minus sign to represent the opposite or additive inverse of an element is something strongly established, as you can see in papers, textbooks and programming languages (e.g., a quick reference: http://mathworld.wolfram.com/AdditiveInverse.html). 

Actually, from an abstract algebra point of view, in a group, _subtraction doesn't exist_. What does exist in a group is addition and the opposite, and subtraction is mere syntactic sugar on top of this. So if we accept that the sign for subtraction is the minus, then it doesn't make sense to use a different symbol to the actual operator underneath.

tests/test_primitives/test_point/test_point_serializers.py
cygnusv: Yep, good idea. Similar code is in other functions (e.g., `Point.unsafe_hash_to_point`), and it's ugly as hell. I just created #197 with that.

umbral/point.py
tuxxy: I'm not necessarily trying to protect the users from anything, but I'm trying to protect the readability of the code. It would be pretty easy to mistake a negation for a subtraction.

Does this mean you're willing to get rid of subtraction and simply add a negation?

michwill: Ah. Removing subtraction while having negation would actually be neat.

Although, I'm not against the original `~`. But if negation is technically more precise, so be it

cygnusv: I didn't mean to get rid of subtraction. As many other times, I tend to push towards following long established conventions. In this case, to use the same sign for the negation and subtraction operators. There's no risk of mistaking anything, they are different types of operators (unary vs binary), with different precedence orders. I simply don't see any value of using two different operators.

tuxxy: Alright, I'm sold.

cygnusv:
> I also noticed that we aren't doing anything to check that the point coordinates are lower than the order of the base field. Is there anything we can do to make that check happen?

We could, but that's something that OpenSSL handles internally. We would be duplicating the check. It's not like when we check the range of CurveBNs with `openssl._bn_is_on_curve`, since in that case we don't allow for zeroes.

tuxxy:
I would venture to say that it's more of a design to check it for defense-in-depth, in that case.

cygnusv:
I'm afraid there's no OpenSSL function to get the order of the base field. It's still possible to inspect the `ec_group_st` struct to get such [modulo](https://github.com/openssl/openssl/blob/master/crypto/ec/ec_lcl.h#L220), but I don't feel comfortable doing that from Python. Plus, I don't see much benefit since ultimately all operations in OpenSSL are performed according to such modulo, and I'm not aware that there exists attacks that exploit coordinates outside the range.

PR #201, opened 2018-07-20
Improve (de)serialization of CurveBN and Points

cygnusv:
## What this does ##
This PR aims to improve the serialization and deserialization of CurveBNs and Points, by using native OpenSSL functions when possible.

Previously, Python `ints` were used as intermediate representation during serialization and deserialization of CurveBNs and Points. This has been bothering @tuxxy and I for a long time due to potential timing attacks, but we never had the time to solve this. Similar concerns have been raise recently in cryptography.io (https://github.com/pyca/cryptography/issues/4235).

**Note**: Depends on #204


jMyles:
Without having done a thorough walk through to assess the timing characteristics, I'm initially given pause by the idea that pyUmbral is the right place for a canonical function to convert an OpenSSL BigNum to bytes and back again.  Is this really our job?  Is nobody else maintaining such simple functionality in a safe way?

tuxxy:
This looks great. Better to keep it here for the time being. Make a PR there whenever you want. When (if) it gets included in a new release, then we can deprecate the functions here.

cygnusv:
I just reduced the scope of this PR to only `openssl._bytes_to_bn`. I'm having problems with the other direction (`openssl._bn_to_bytes`), so I prefer to leave it like this for the moment.

tests/test_keys/test_key_fragments.py
```diff
@@ -30,18 +30,24 @@ def test_kfrag_serialization(alices_keys, bobs_keys):
...
-    assert new_frag._bn_key == kfrags[0]._bn_key
-    assert new_frag._point_noninteractive == kfrags[0]._point_noninteractive
-    assert new_frag._point_commitment == kfrags[0]._point_commitment
-    assert new_frag._point_xcoord == kfrags[0]._point_xcoord
+
+    for kfrag in kfrags:
+        kfrag_bytes = kfrag.to_bytes()
+        assert len(kfrag_bytes) == KFrag.expected_bytes_length(curve)
+
+        new_kfrag = KFrag.from_bytes(kfrag_bytes)
```
tuxxy: :+1:

tests/test_primitives/test_point/test_point_serializers.py
```diff
@@ -144,19 +143,18 @@ def test_serialize_point_at_infinity():
...
     point_at_infinity = p - p
     
-    with pytest.raises(InternalError) as e:
-        _bytes_point_at_infinity = point_at_infinity.to_bytes()
-
-    # We want to catch specific InternalExceptions:
-    # - Point at infinity (code 106)
-    assert e.value.err_code[0].reason == 106
+    bytes_point_at_infinity = point_at_infinity.to_bytes()
+    assert bytes_point_at_infinity == b'\x00'
```
tuxxy: Seems pretty endless.

tests/test_primitives/test_point/test_point_serializers.py
```diff
@@ -144,19 +143,18 @@ def test_serialize_point_at_infinity():
...
 
-    # Testing that a point with the x coordinate greater than the curve order is still valid
-    compressed = 0x02fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2c
-    compressed = compressed.to_bytes(32+1, byteorder='big')
+    # Testing that a point with x coordinate greater than the curve order is still valid.
+    # In particular, we will test the last valid point from the default curve (secp256k1)
+    # whose x coordinate is `field_order - 3` and is greater than the order of the curve
+
+    field_order = 2**256 - 0x1000003D1
+    compressed = b'\x02' + (field_order-3).to_bytes(32, 'big')
```
tuxxy: Real good test. :+1:

umbral/curve.py
```diff
@@ -98,6 +99,13 @@ def field_order_size_in_bytes(self) -> int:
             self.__field_order_size_in_bytes = (size_in_bits + 7) // 8
         return self.__field_order_size_in_bytes
 
+    @property
+    def group_order_size_in_bytes(self) -> int:
+        if self.__group_order_size_in_bytes is None:
```
tuxxy: Why are you checking this here? Python handles the caching when you do `@property`.

umbral/openssl.py
```diff
@@ -121,6 +121,39 @@ def _int_to_bn(py_int: int, curve: 'Curve'=None, set_consttime_flag=True):
...
+    If set_consttime_flag is set to True, OpenSSL will use constant time
+    operations when using this BIGNUM.
+    """
+    bn = _get_new_BN(set_consttime_flag)
+    backend._lib.BN_bin2bn(bytes_seq, len(bytes_seq), bn)
+    backend.openssl_assert(bn != backend._ffi.NULL)    
+    return bn
+
+@typing.no_type_check
+def _bn_to_bytes(bignum, length : int = None):
```
tuxxy: Cool :+1:

umbral/curve.py
cygnusv: It seems it doesn't. https://stackoverflow.com/questions/6428723/python-are-property-fields-being-cached-automatically

tests/test_keys/test_key_fragments.py
```diff
@@ -30,18 +30,24 @@ def test_kfrag_serialization(alices_keys, bobs_keys):
...
-    assert len(kfrag_bytes) == KFrag.expected_bytes_length(curve)
 
-    new_frag = pre.KFrag.from_bytes(kfrag_bytes)
-    assert new_frag._id == kfrags[0]._id
-    assert new_frag._bn_key == kfrags[0]._bn_key
-    assert new_frag._point_noninteractive == kfrags[0]._point_noninteractive
-    assert new_frag._point_commitment == kfrags[0]._point_commitment
-    assert new_frag._point_xcoord == kfrags[0]._point_xcoord
+    for kfrag in kfrags:
+        kfrag_bytes = kfrag.to_bytes()
```
KPrasch: This still works as `bytes(kfrag)`, also?

tests/test_primitives/test_curvebn/test_curvebn_serializers.py
```diff
@@ -0,0 +1,45 @@
...
+def test_invalid_deserialization(curve):
+	size_in_bytes = CurveBN.expected_bytes_length(curve)
+	
+	# All-zeros bytestring are invalid (i.e., 0 < bn < order of the curve)
+	zero_bytes = bytes(size_in_bytes)
+	with pytest.raises(ValueError):
+		_bn = CurveBN.from_bytes(zero_bytes, curve)
+
+	# All-ones bytestring is invalid too (since it's greater than order)
+	lots_of_ones = 2**(8*size_in_bytes) - 1
```
KPrasch: Two can be as bad as one

tests/test_primitives/test_point/test_point_serializers.py
```diff
@@ -67,15 +67,15 @@ def test_bytes_serializers(point_bytes, nid, curve):
     for point_representation in representations:
 
         malformed_point_bytes = point_representation + b'0x'
-        with pytest.raises(ValueError):
+        with pytest.raises(InternalError):
```
KPrasch: ....as opposed to an external one?

umbral/openssl.py
KPrasch: Hmm - unless I'm missing something this seems to break existing serialization conventions.  Why isn't this `Bignum.from_bytes` (perhaps it is invoked there)?

tests/test_primitives/test_point/test_point_serializers.py
cygnusv: `ValueError` was raised by our serialization logic, while `InternalError` is what cryptography.io raises to account for OpenSSL errors.

umbral/openssl.py
cygnusv: Yep, this is invoked in `CurveBN.to_bytes()`

PR #204, opened 2018-07-26
Detect invalid KFrags before use

cygnusv:
This PR enforces KFrag validation before re-encryption. Since KFrag validation requires knowledge of the correctness keys, and these can be obtained from the capsule, now Ursula has to set the correctness keys in the capsule before re-encryption. 

**Motivation:** Working on new stuff for #201, I introduced a serialization bug by mistake which went mostly undetected by the test suite. The bug was probabilistic, so it took me a lot of time to kill it. Apart from the fact that there are weak spots in the tests, the most alarming problem for me is that it was possible to deserialize invalid KFrags and use them to perform re-encryption and correctness proofs, which of course were invalid. This can be detected at decryption time, but of course this is not desirable, and could lead to a lot of problems for Ursulas.

Although this concern has already been raised for the NuCypher network (https://github.com/nucypher/nucypher/issues/167), I feel we should do some in-depth defense here and enforce this too at the pyUmbral level. 

**Other stuff:** This PR removes `UmbralPublicKey.get_pubkey()` method.

### DIFF
--- a/docs/notebooks/pyUmbral Simple API.ipynb
+++ b/docs/notebooks/pyUmbral Simple API.ipynb
@@ -203,7 +203,7 @@
    "source": [
     "\n",
     "## Ursulas Re-encrypt; Bob attaches fragments to capsule\n",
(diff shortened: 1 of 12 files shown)


tuxxy:
What's left in this PR?

PR #215, opened 2018-08-31
Some __repr__ and input validation

cygnusv:
A really small PR with some pending stuff:

- Implements `__repr__` for `CapsuleFrag`, `Capsule`, `UmbralPublicKey`, and `KFrag`
- Input validation on `pre.decrypt` and `pre.reencrypt`

### DIFF
--- a/umbral/fragments.py
+++ b/umbral/fragments.py
@@ -117,6 +117,9 @@ class KFrag(object):
     def __hash__(self):
         return hash(bytes(self._id))
 
+    def __repr__(self):
+        return "{}:{}".format(self.__class__.__name__, self._id.hex()[:15])
+
 
 class CorrectnessProof(object):
     def __init__(self, point_e2: Point, point_v2: Point, point_kfrag_commitment: Point,
@@ -277,3 +280,6 @@ class CapsuleFrag(object):
 
     def __bytes__(self) -> bytes:
         return self.to_bytes()
+
+    def __repr__(self):
+        return "CFrag:{}".format(self._point_e1.to_bytes().hex()[2:17])
--- a/umbral/pre.py
+++ b/umbral/pre.py
@@ -29,7 +29,7 @@ from bytestring_splitter import BytestringSplitter
 from umbral._pre import prove_cfrag_correctness
 from umbral.config import default_curve
 from umbral.curvebn import CurveBN
-from umbral.dem import UmbralDEM, DEM_KEYSIZE
+from umbral.dem import UmbralDEM, DEM_KEYSIZE, DEM_NONCE_SIZE
 from umbral.fragments import KFrag, CapsuleFrag
 from umbral.keys import UmbralPrivateKey, UmbralPublicKey
 from umbral.params import UmbralParameters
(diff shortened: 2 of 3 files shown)


umbral/pre.py
```diff
@@ -278,7 +278,7 @@ def __eq__(self, other: 'Capsule') -> bool:
             # This is not constant time obviously, but it's hard to imagine how this is valuable as
             # an attacker already knows about her own Capsule.  It's possible that a Bob, having
             # activated a Capsule, will make it available for comparison via an API amidst other
-            # (dormat) Capsules.  Then an attacker can, by alternating between activated and dormant
```
KPrasch: Excellent.

umbral/pre.py
```diff
@@ -295,6 +295,8 @@ def __hash__(self) -> int:
     def __len__(self) -> int:
         return len(self._attached_cfrags)
 
+    def __repr__(self):
+        return "{}:{}".format(self.__class__.__name__, hex(int(self._bn_sig))[2:17])
```
KPrasch: Why strip the first 2?

cygnusv: Because it's the 0x thingy.

umbral/fragments.py
```diff
@@ -277,3 +280,6 @@ def attach_proof(self, e2: Point, v2: Point, u1: Point, u2: Point, z3: CurveBN,
 
     def __bytes__(self) -> bytes:
         return self.to_bytes()
+
+    def __repr__(self):
+        return "CFrag:{}".format(self._point_e1.to_bytes().hex()[2:17])
```
KPrasch: Missing a newline here

umbral/pre.py
```diff
@@ -380,10 +382,10 @@ def split_rekey(delegating_privkey: UmbralPrivateKey, signer: Signer,
 def reencrypt(kfrag: KFrag, capsule: Capsule, provide_proof: bool = True, 
               metadata: Optional[bytes] = None) -> CapsuleFrag:
 
-    if not capsule.verify():
+    if capsule is None or not capsule.verify():
```
KPrasch: This is essentially the "last second" to check the `capsule is None`, right?

umbral/pre.py
```diff
@@ -524,22 +526,21 @@ def decrypt(ciphertext: bytes, capsule: Capsule, decrypting_key: UmbralPrivateKe
     and return the resulting cleartext.
     """
 
+    if not isinstance(ciphertext, bytes) or len(ciphertext) < DEM_NONCE_SIZE:
+        raise ValueError("Input ciphertext must be a bytes object of length >= {}".format(DEM_NONCE_SIZE))
```
tuxxy: I think I would leave out the check to see if it is `bytes` and only check length.

Also, just nitpicking, but change "bytes object" to "bytestring", but it makes sense as it is because Python. :P

umbral/pre.py
```diff
@@ -524,22 +526,21 @@ def decrypt(ciphertext: bytes, capsule: Capsule, decrypting_key: UmbralPrivateKe
...
         # (i.e., this is an original capsule)
-        decapsulated_key = _decapsulate_original(decrypting_key, capsule)
-        dem = UmbralDEM(decapsulated_key)
-
+        encapsulated_key = _decapsulate_original(decrypting_key, capsule)
         capsule_bytes = bytes(capsule)
-        cleartext = dem.decrypt(ciphertext, authenticated_data=capsule_bytes)
+
+    dem = UmbralDEM(encapsulated_key)
+    cleartext = dem.decrypt(ciphertext, authenticated_data=capsule_bytes)
```
tuxxy: This reminds me that we still need to add cipher modularity for this, too.

PR #218, opened 2018-09-16
[WIP] KFrag validity no longer requires knowledge of delegating and receiving public keys

cygnusv:
The goal of this PR is to provide the minimum update that allows NuCypher to synchronize with current state of pyUmbral. In particular, knowledge of delegating and receiving public keys is now optional to KFrag validity; this is decided by Alice when generating the kfrags using the new flags `sign_delegating_key` and `sign_receiving_key`. 

This PR bumps pyUmbral's version: `0.1.0-alpha.4 → 0.1.1-alpha.0`

### DIFF
--- a/umbral/_pre.py
+++ b/umbral/_pre.py
@@ -152,7 +152,7 @@ def verify_kfrag(kfrag: 'KFrag',
 
     u = params.u
 
-    id = kfrag._id
+    kfrag_id = kfrag._id
     key = kfrag._bn_key
     u1 = kfrag._point_commitment
     ni = kfrag._point_noninteractive
@@ -169,7 +169,7 @@ def verify_kfrag(kfrag: 'KFrag',
     if receiving_pubkey is None:
         receiving_pubkey = b'\x00' * pubkey_size
 
-    validity_input = (id, delegating_pubkey, receiving_pubkey, u1, ni, xcoord)
+    validity_input = (kfrag_id, delegating_pubkey, receiving_pubkey, u1, ni, xcoord)
 
     kfrag_validity_message = bytes().join(bytes(item) for item in validity_input)
     valid_kfrag_signature = kfrag.signature.verify(kfrag_validity_message, signing_pubkey)
--- a/umbral/_pre.py
+++ b/umbral/_pre.py
@@ -31,7 +31,7 @@ def prove_cfrag_correctness(cfrag: 'CapsuleFrag',
                             metadata: Optional[bytes] = None
                             ) -> None:
 
-    params = capsule._umbral_params
+    params = capsule.params
 
     # Check correctness of original ciphertext
     if not capsule.verify():
@@ -63,7 +63,7 @@ def prove_cfrag_correctness(cfrag: 'CapsuleFrag',
 
     z3 = t + h * rk
 
-    cfrag.attach_proof(e2, v2, u1, u2, metadata=metadata, z3=z3, kfrag_signature=kfrag.signature)
+    cfrag.attach_proof(e2, v2, u1, u2, metadata=metadata, z3=z3, kfrag_signature=kfrag.signature_for_bob)
 
 
 def assess_cfrag_correctness(cfrag: 'CapsuleFrag', capsule: 'Capsule') -> bool:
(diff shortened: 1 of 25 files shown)


umbral/_pre.py
```diff
@@ -113,8 +108,15 @@ def assess_cfrag_correctness(cfrag: 'CapsuleFrag', capsule: 'Capsule') -> bool:
     xcoord = cfrag._point_xcoord
     kfrag_id = cfrag._kfrag_id
 
-    kfrag_validity_message = bytes().join(
-        bytes(material) for material in (kfrag_id, delegating_point, receiving_point, u1, ni, xcoord))
+    if delegating_pubkey is None:
+        delegating_pubkey = bytes(UmbralPublicKey.expected_bytes_length())
+
+    if receiving_pubkey is None:
+        receiving_pubkey = bytes(UmbralPublicKey.expected_bytes_length())
```
tuxxy: While this totally works, I hate using it because it makes no sense.

Instead, I prefer `b'\x00' * UmbralPublicKey.expected_bytes_length()`

cygnusv: Yeah, I agree it makes no sense, while your approach is crystalline. I'll make the changes.

umbral/pre.py
```diff
@@ -363,8 +368,15 @@ def split_rekey(delegating_privkey: UmbralPrivateKey, signer: Signer,
...
 
-        kfrag_validity_message = bytes().join(
-            bytes(material) for material in (id, pubkey_a_point, pubkey_b_point, u1, ni, xcoord))
+        pubkey_size = UmbralPublicKey.expected_bytes_length(curve=params.curve)
+        if not sign_delegating_key:
+            delegating_pubkey = b'\x00' * pubkey_size
+        if not sign_receiving_key:
+            receiving_pubkey = b'\x00' * pubkey_size
+
+        validity_input = (id, delegating_pubkey, receiving_pubkey, u1, ni, xcoord)
```
KPrasch: Perhaps nit-picking here but let's be careful with name shadowing the builtin functions (`id` in this case), It can lead to some unnecessary code fragility.

umbral/pre.py
```diff
@@ -331,7 +336,7 @@ def split_rekey(delegating_privkey: UmbralPrivateKey, signer: Signer,
     ni = priv_ni * g
     d = CurveBN.hash(ni, pubkey_b_point, pubkey_b_point * priv_ni, params=params)
 
-    coeffs = [privkey_a_bn * (~d)]
+    coeffs = [delegating_privkey * (~d)]
```
KPrasch: Phew - Thank you! :-)

KPrasch:
> In particular, knowledge of delegating and receiving public keys is now optional to KFrag validity; this is  decided by Alice when generating the kfrags using the new flags sign_delegating_key and  sign_receiving_key."

@cygnusv - We need to start a changelog!

umbral/_pre.py
```diff
@@ -131,22 +135,24 @@ def assess_cfrag_correctness(cfrag: 'CapsuleFrag', capsule: 'Capsule') -> bool:
...
-    u = params.u
+    if receiving_pubkey and receiving_pubkey.params != params:
+            raise ValueError("The receiving key uses different UmbralParameters")
 
-    delegating_point = delegating_pubkey.point_key
-    receiving_point = receiving_pubkey.point_key
+    u = params.u
 
-    id = kfrag._id
+    kfrag_id = kfrag._id
```
jMyles: Certainly a step in the right direction.  :-)

But again: doesn't this suggest that `kfrag._id` wants to be public?

umbral/_pre.py
```diff
@@ -155,8 +161,17 @@ def verify_kfrag(kfrag: 'KFrag',
     #  We check that the commitment u1 is well-formed
     correct_commitment = u1 == key * u
 
-    kfrag_validity_message = bytes().join(
-        bytes(material) for material in (id, delegating_point, receiving_point, u1, ni, xcoord))
+    pubkey_size = UmbralPublicKey.expected_bytes_length(curve=params.curve)
+
+    if delegating_pubkey is None:
```
jMyles: I think I need a quick explanation of this - we use zeroed-out keys if none are provided?  I'll just ask on Discord.

umbral/_pre.py
cygnusv: In a way, yes ... although I kind of recall a consensus on that "_this is fine as long as we are calling private attributes from a private module_" (in this case`umbral._pre`). My memory can be wrong on this, though.

The problem is that I've just checked and there are several private attributes of `KFrag` (including `_id`) that are called in `umbral.pre`, so I guess I'll have to make these public. But I think these PR may not be the place, because it's not only `KFrag`, this is happening all over the place. 

And by the way, perhaps unrelated, I don't like the `umbral._pre` module. Ideally, I'd want to remove it once we assess the situation with the right attributes visibility.

Issue #122, opened 2018-04-19
Refactor code for password-protected (de)serialization in UmbralPrivateKey and UmbralKeyingMaterial


cygnusv:
There is currently a lot of common code in the password-protected (de)serialization of `UmbralPrivateKey` and `UmbralKeyingMaterial`. In PR#111 I tried at first to do it through a common abstract class, although we ended up rejecting this option. I agree that an abstract parent class is too much just for this thing. But we should reuse this code in both classes, for example, through a function in module `umbral.keys`

PR #219, opened 2018-09-19
Random oracles module and code reorganization

cygnusv:
The main goal of this PR is introducing a new module for the random oracles used by Umbral. This will imply a new release of pyumbral, which is necessary for a forthcoming NuCypher release (see the [`vodka` branch](https://github.com/nucypher/nucypher/pull/507)).

### What this does 
- Reorganization and improvement of all the random oracles used by Umbral (i.e, the KDF, `hash_to_point`, `hash_to_curvebn`). 
- Hash functions now allow to define a 64-bytes long `customization_string`, intended for supporting separation of application domains. This customization string is prepended to the data fed to the hash function. By default, this string is all zeroes; if the string used is shorter than 64 bytes, it is right padded with the byte `0x00`.
- Includes a new 64-byte hash function called `ExtendedKeccak`, as an alternative to Blake2b, to facilitate integration of ZK proofs verification by Ethereum smart contracts. The result of `ExtendedKeccak(data)` is simply `Keccak(0x00||custom_str||data)||Keccak(0x01||custom_str||data)`. This hash function is only used internally by the ZKP, as running Blake2b by a smart contract is currently very expensive, while Keccak is native to the EVM.
- Important reorganization and improvement of code: 
    - `umbral.fragments` (which contained `KFrag` and `CapsuleFrag`) disappears, and is 
    - `umbral._pre` disappears. It was causing unnecessary circular dependencies.
    - Fixes scope of attributes through all the codebase
    - Brings mypy checks back and solves all typing errors!
- Improved key wrapping for `UmbralPrivateKey` and `UmbralKeyingMaterial` (Solves #122). This will simplify NuCypher's Keyring in the future. In addition, it supports adding other password-based key derivation methods in the future as discussed (in case we want to include Argon2). Default method continues to be Scrypt.


tests/metrics/reencryption_firehose.py
```diff
@@ -23,7 +23,7 @@
 
 sys.path.append(os.path.abspath(os.getcwd()))
 
-from umbral.fragments import KFrag
+from umbral.kfrags import KFrag
```
tuxxy: Why did you split fragments up? Naming from `kfrags.KFrag` and `cfrags.CapsuleFrag` seems really redundant.

umbral/keys.py
```diff
@@ -313,76 +401,64 @@ def derive_privkey_by_label(self,
 
         key_material = HKDF(
             algorithm=hashes.BLAKE2b(64),
-            length=64,
+            length=64,  # TODO: Generalize for multicurve
```
tuxxy: What exactly are you wanting to generalize for multicurve support? This is just deriving a key, then hashing it using whatever curve params were chosen in the call to `hash_to_curvebn` below.

umbral/kfrags.py
```diff
@@ -0,0 +1,192 @@
...
+from umbral.config import default_curve, default_params
+from umbral.curvebn import CurveBN
+from umbral.keys import UmbralPublicKey
+from umbral.point import Point
+from umbral.signing import Signature
+from umbral.params import UmbralParameters
+from umbral.curve import Curve
+
+from constant_sorrow.constants import NO_KEY, DELEGATING_ONLY, RECEIVING_ONLY, DELEGATING_AND_RECEIVING
+
```
tuxxy: PEP8 nit

```suggestion


```

tests/metrics/reencryption_firehose.py
cygnusv: It never had sense to group KFrags and CFrags together just because they were "fragments". It also created some unnecessary circular imports, which made difficult to solve type errors found by mypy.

umbral/keys.py
cygnusv: You're right! I'll remove the TODO.

tests/metrics/reencryption_firehose.py
tuxxy: Agreed, but I think a better name exists other than `kfrags` and `cfrags`... I'll think about it.

cygnusv: Sure, I guess the module name can change if there's a better alternative.

umbral/keys.py
cygnusv: ✅

umbral/keys.py
```diff
@@ -69,7 +69,7 @@ def __call__(self,
                 backend=default_backend()
             ).derive(password)
         except InternalError as e:
-            required_memory = 128 * 2**_scrypt_cost * 8 // (1024**1024)
+            required_memory = 128 * 2**_scrypt_cost * 8 // (10**6)
```
tuxxy:
```suggestion
required_memory = 128 * 2**_scrypt_cost * 8 // (1024**2)```

Comment (tuxxy) [reply umbral/keys.py]: This is more accurate. :)

File: tests/unit/test_capsule_serializers.py [hunk 5]
```diff
@@ -42,26 +42,24 @@ def test_capsule_serialization(capsule):
...
+                                                     point_e=Point.gen_rand(),
+                                                     point_v=42,
+                                                     bn_sig=CurveBN.gen_rand())
 
     with pytest.raises(TypeError):
-        capsule_of_questionable_parentage = pre.Capsule(params,
-                                                        point_e=Point.gen_rand(),
-                                                        point_v=Point.gen_rand(),
-                                                        bn_sig=42)
+        _capsule_of_questionable_parentage = Capsule(params,
```
KPrasch: oh my

umbral/keys.py
tuxxy: Or just remove it, I have a PR on cryptography.io cooking up lol

cygnusv: Oh yeah! as soon as that's merged, this exception can be deprecated.

tuxxy: Yeah, so maybe just undo this lol

umbral/keys.py
```diff
@@ -33,9 +34,119 @@
...
+             **kwargs) -> bytes:
+    """
+    Wraps a key using a provided wrapping key. Alternatively, it can derive
+    the wrapping key from a password.
+    :param key_to_wrap:
+    :param wrapping_key:
+    :param password:
+    :return:
+    """
+    if all((password, wrapping_key)) or not any((password, wrapping_key)):
```
KPrasch: A bit wordy - perhaps:

```python
if bool(password) ^ bool(wrapping_key):
    raise ValueError("Either password or wrapping_key must be passed; Not both.")
```

umbral/keys.py
```diff
@@ -58,82 +169,60 @@ def gen_key(cls, params: Optional[UmbralParameters] = None) -> 'UmbralPrivateKey
...
-                n=2**_scrypt_cost,
-                r=8,
-                p=1,
-                backend=default_backend()
-            ).derive(password)
+        key_bytes = self.bn_key.to_bytes()
 
-            umbral_privkey = SecretBox(key).encrypt(umbral_privkey)
-            umbral_privkey += salt
+        if any((wrapping_key, password)):
```
KPrasch: Why use `any` here?

umbral/keys.py
cygnusv: wonderful!

umbral/pre.py
```diff
@@ -375,45 +417,38 @@ def _decapsulate_reencrypted(receiving_privkey: UmbralPrivateKey, capsule: Capsu
...
-            lambda_i = lambda_coeff(x, xs)
-            e_summands.append(lambda_i * cfrag._point_e1)
-            v_summands.append(lambda_i * cfrag._point_v1)
-
-        e_prime = sum(e_summands[1:], e_summands[0])
-        v_prime = sum(v_summands[1:], v_summands[0])
-    else:
-        e_prime = capsule._attached_cfrags[0]._point_e1
-        v_prime = capsule._attached_cfrags[0]._point_v1
+    xs = [hash_to_curvebn(precursor,
```
KPrasch: I actually think a formal loop would be more readable here.

umbral/random_oracles.py
```diff
@@ -0,0 +1,216 @@
...
+import sha3
+
+from umbral import openssl
+from umbral.curvebn import CurveBN
+from umbral.point import Point
+from umbral.params import UmbralParameters
+from umbral.config import default_params
+
+
+class Hash(ABC):
```
KPrasch: Wow - not messing around here

umbral/keys.py
cygnusv: Because I want to call `wrap_key` only when `wrapping_key is not None or password is not None`. I guess `any((wrapping_key, password))` is shorter, but perhaps less clear?

KPrasch: I was thinking 

```python
if wrapping_key or password:
```

cygnusv: ... yep.

.circleci/config.yml
```diff
@@ -220,10 +220,10 @@ jobs:
           name: Install dependencies
           command: pipenv sync --three --dev
       - run:
-          name: Run Mypy Static Type Checks (Always Succeed)
+          name: Run Mypy Static Type Checks
           command: |
               pipenv install --dev --skip-lock lxml
-              pipenv run mypy --xslt-html-report ./mypy_report ./umbral --config-file=mypy.ini || true
+              pipenv run mypy --xslt-html-report ./mypy_report ./umbral --config-file=mypy.ini
```
KPrasch: :zap:

umbral/kfrags.py
```diff
@@ -0,0 +1,192 @@
...
+NO_KEY(b'\x00')
```
KPrasch: Let's stop using constant sorrow for this purpose, no?

cygnusv: Why is that?

umbral/keys.py
cygnusv: ✅

umbral/keys.py
cygnusv: ✅

umbral/pre.py
cygnusv: ✅

umbral/kfrags.py
cygnusv: I made the change, it's simpler now, so I guess that's why you suggested it.

umbral/kfrags.py
```diff
@@ -0,0 +1,192 @@
...
+DELEGATING_ONLY(b'\x01')
+RECEIVING_ONLY(b'\x02')
+DELEGATING_AND_RECEIVING(b'\x03')
+
```
jMyles: Why are you representing each of these as one byte?  Is this just for space-saving?

PR #220, opened 2018-10-06
Improved verification of KFrags

cygnusv:
The original goal of this PR is to update pyUmbral so that it allows NuCypher to be in sync. In particular, knowledge of delegating and receiving public keys is now optional to KFrag validity; this is decided by Alice when generating the kfrags using the new flags `sign_delegating_key` and `sign_receiving_key`. Apart from this, there are many improvements in this PR, like for example, removing the activated capsule state, or the test automation for pyUmbral's Jupyter notebook.

This PR is based on #218, which I closed since there were a lot of changes since then.

More detailed description of the changes:
- [x] KFrags have now two signatures (one to be verified by the proxy and another by Bob)
- [x] Correctness keys are optional in the signature for the proxy. 
- [x] API change: `pre.split_rekey` --> `pre.generate_kfrags`
- [x] Remove activated capsule state from `Capsule`
- [x] Update Jupyter notebook and automate testing using pytest's plugin nbval
- [x] Update README, docs, and scripts

**TODO** (once PR is approved):
- [ ] Regenerate test vectors
- [ ] Bump version

### DIFF
--- a/docs/examples/umbral_simple_api.py
+++ b/docs/examples/umbral_simple_api.py
@@ -17,14 +17,14 @@ You should have received a copy of the GNU General Public License
 along with pyUmbral. If not, see <https://www.gnu.org/licenses/>.
 """
 
-# 1
+#1
 # Sets a default curve (secp256k1)
 import random
 from umbral import pre, keys, config, signing
 
 config.set_default_curve()
 
-# 2
+#2
 # Generate an Umbral key pair
 # ---------------------------
 # First, Let's generate two asymmetric key pairs for Alice:
@@ -37,7 +37,7 @@ alices_signing_key = keys.UmbralPrivateKey.gen_key()
 alices_verifying_key = alices_signing_key.get_pubkey()
 alices_signer = signing.Signer(private_key=alices_signing_key)
 
-# 3
+#3
 # Encrypt some data for Alice
 # ---------------------------
 # Now let's encrypt data with Alice's public key.
@@ -49,7 +49,7 @@ plaintext = b'Proxy Re-encryption is cool!'
 ciphertext, capsule = pre.encrypt(alices_public_key, plaintext)
 print(ciphertext)
 
-# 4
+#4
 # Decrypt data for Alice
 # ----------------------
 # Since data was encrypted with Alice's public key,
@@ -60,18 +60,18 @@ cleartext = pre.decrypt(ciphertext=ciphertext,
                         decrypting_key=alices_private_key)
 print(cleartext)
 
-# 5
+#5
 # Bob Exists
 # -----------
 
 bobs_private_key = keys.UmbralPrivateKey.gen_key()
 bobs_public_key = bobs_private_key.get_pubkey()
 
-# 6
+#6
 # Bob receives a capsule through a side channel (s3, ipfs, Google cloud, etc)
 bob_capsule = capsule
 
-# 7
+#7
 # Attempt Bob's decryption (fail)
 try:
     fail_decrypted_data = pre.decrypt(ciphertext=ciphertext,
@@ -80,7 +80,7 @@ try:
 except:
     print("Decryption failed! Bob doesn't has access granted yet.")
 
-# 8
+#8
 # Alice grants access to Bob by generating kfrags 
 # -----------------------------------------------
 # When Alice wants to grant Bob access to open her encrypted messages, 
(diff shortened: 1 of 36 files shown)


docs/examples/umbral_simple_api.py
```diff
@@ -17,14 +17,14 @@
 along with pyUmbral. If not, see <https://www.gnu.org/licenses/>.
 """
 
-#1
+# 1
```
tuxxy: Can you remove these spaces? It makes demos more difficult cause you have to do the space and when you have one hand and the other has a mic, it's a bit annoying. :P

docs/examples/umbral_simple_api.py
```diff
@@ -49,39 +49,38 @@
...
-#7
+# 7
 # Attempt Bob's decryption (fail)
 try:
-    fail_decrypted_data = pre.decrypt(ciphertext=ciphertext, 
-    								  capsule=bob_capsule, 
-    								  decrypting_key=bobs_private_key)
+    fail_decrypted_data = pre.decrypt(ciphertext=ciphertext,
+                                      capsule=bob_capsule,
+                                      decrypting_key=bobs_private_key)
```
tuxxy: 👍

umbral/fragments.py
```diff
@@ -57,7 +73,15 @@ def expected_bytes_length(cls, curve: Optional[EllipticCurve] = None) -> int:
...
         point_size = Point.expected_bytes_length(curve)
 
-        return (bn_size * 4) + (point_size * 3)
+        # self.id --> 1 bn_size
+        # self._bn_key --> 1 bn_size
+        # self._point_commitment --> 1 point_size
+        # self._point_precursor --> 1 point_size
+        # self.signature_for_proxy --> 2 bn_size
+        # self.signature_for_bob --> 2 bn_size
+        # self.keys_in_signature --> 1
```
tuxxy: I think this is overly-verbose, but should definitely have a comment for `keys_in_signature`. The `bn_size` and `point_size` stuff is pretty self-evident in the `return`.

umbral/fragments.py
```diff
@@ -267,8 +313,15 @@ def to_bytes(self) -> bytes:
     def verify_correctness(self, capsule: 'Capsule') -> bool:
         return assess_cfrag_correctness(self, capsule)
 
-    def attach_proof(self, e2: Point, v2: Point, u1: Point, u2: Point, z3: CurveBN, kfrag_signature: Signature,
+    def attach_proof(self,
```
tuxxy: Is this a weird github thing or is there a bunch of whitespace after this?

umbral/pre.py
```diff
@@ -51,46 +48,34 @@ def __init__(self, message: str, offending_cfrags: List[CapsuleFrag]) -> None:
...
-                 bn_sig: Optional[CurveBN] = None,
-                 point_e_prime: Optional[Point] = None,
-                 point_v_prime: Optional[Point] = None,
-                 point_noninteractive: Optional[Point] = None,
-                 delegating_pubkey: Optional[UmbralPublicKey] = None,
-                 receiving_pubkey: Optional[UmbralPublicKey] = None,
-                 verifying_pubkey: None = None
+                 point_e: Point,
+                 point_v: Point,
+                 bn_sig: CurveBN,
```
tuxxy: 👍

umbral/pre.py
```diff
@@ -51,46 +48,34 @@ def __init__(self, message: str, offending_cfrags: List[CapsuleFrag]) -> None:
...
+        }   # type: dict
+
+    class NotValid(ValueError):
+        """
+        raised if the capsule does not pass verification.
+        """
 
     @classmethod
-    def expected_bytes_length(cls, curve: Optional[EllipticCurve] = None, activated: bool = False) -> int:
+    def expected_bytes_length(cls, curve: Optional[Curve] = None) -> int:
```
tuxxy: 👍

umbral/fragments.py
cygnusv: Well, apart from `keys_in_signature`, the size of `self.id` is arbitrary too, since it's a byte string. I agree that it's a big comment, but in this case I think it's justified.

KPrasch:
@cygnusv Looks like using `constant-sorrow` from PyPI causes a python 3.5 failure?

PR #231, opened 2018-12-12
Change signature's internal hash algorithm from BLAKE2B to SHA256

cygnusv:
As part of the on-chain verification of re-encryption in the `vodka` branch of nucypher, we need to check some signatures by Alice, Bob and Ursula. Unfortunately, umbral uses Blake2b as the internal hash algorithm, which is quite inconvenient to compute in a smart contract. Setting aside that ECDSA + Blake2b is a weird combination, something like ECDSA + SHA256 is supported natively by the EVM, and it's a de-facto standard. 

Note that this is backwards-incompatible.

After this is merged, we'll have a new release (**v0.1.3a0**).


vectors/vectors_kfrags.json
```diff
@@ -1,40 +1,40 @@
...
     }
-  ],
-  "verifying_key": "03203a4d8e286d6fb5075657afda918d071e1e82f80ca05b96bdb686055963779a",
-  "params": "default",
-  "receiving_key": "02464902352f6aa7b23d85a3f1f62868989910ded3dc5ada6fc5aabe17757482da",
-  "description": "This is a collection of KFrags generated under the enclosed delegating, verifying and receiving keys. Each of them must deserialize correctly and the call to verify() must succeed.",
-  "name": "Test vectors for KFrags",
-  "delegating_key": "02401e952ba01155935c6c991c0a503b549a7d965a8c6d1eb406f61fc138b775d1"
+  ]
 }
```
KPrasch: Unsure if missing newlines matter in vectors

PR #238, opened 2019-02-19
Update to the README and Docs

cygnusv:
The main change is updating the first paragraph of the README to this:

> pyUmbral is the reference implementation of the Umbral threshold proxy re-encryption scheme. It is open-source, built with Python, and uses OpenSSL and Cryptography.io.
> 
> Using Umbral, Alice (the data owner) can delegate decryption rights to Bob for any ciphertext intended to her, through a re-encryption process performed by a set of semi-trusted proxies or Ursulas. When a threshold of these proxies participate by performing re-encryption, Bob is able to combine these independent re-encryptions and decrypt the original message using his private key.
> 
> pyUmbral is the cryptographic engine behind nucypher, a proxy re-encryption network to empower privacy in decentralized systems.

Other changes:
- Updates docs to match the README, as well as some touchups there.
- Set the `sphinx_rtd_theme` in the docs (the same that NuCypher uses).
- Updates copyright notice to 2019 and removes it from individual file headers.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -78,7 +78,7 @@ pygments_style = 'sphinx'
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'alabaster'
+html_theme = 'sphinx_rtd_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -20,8 +20,8 @@
 # -- Project information -----------------------------------------------------
 
 project = 'pyUmbral'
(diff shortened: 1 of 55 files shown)


docs/source/conf.py
```diff
@@ -20,8 +20,8 @@
 # -- Project information -----------------------------------------------------
 
 project = 'pyUmbral'
-copyright = '2018, Michael Egorov, Justin Myles Holmes, David Nuñez, John Pacific, Kieran Prasch'
-author = 'Michael Egorov, Justin Myles Holmes, David Nuñez, John Pacific, Kieran Prasch'
+copyright = u'2018, Michael Egorov, Justin Myles Holmes, David Nuñez, John Pacific, Kieran Prasch'
```
mswilkison: Actually, shouldn't the copyright be '2019, NuCypher'?

cygnusv: Yeah, and we still have the issue of having it on all file headers. Let me add an additional commit to resolve all that.

cygnusv: ✔️

PR #241, opened 2019-03-27
1st Chimney anniversary

cygnusv:
This PR does:
* Raises `UmbralDecryptionError` instead of `cryptography.exceptions.InvalidTag`, with a more informative error message.
* Updates the docs with a specific section for decryption
* Allows serialization of `UmbralPublicKeys` as hex strings

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,8 +28,6 @@ from umbral import pre
 
 set_default_curve(SECP256K1)
 
-MockKeyPair = namedtuple('TestKeyPair', 'priv pub')
-
 parameters = (
     # (N, M)
     (1, 1),
@@ -56,18 +54,19 @@ kfrag_signing_modes = (
     (True, True), (True, False), (False, True), (False, False)
 )
 
-@pytest.fixture(scope='function')
+
+@pytest.fixture
 def alices_keys():
     delegating_priv = keys.UmbralPrivateKey.gen_key()
     signing_priv = keys.UmbralPrivateKey.gen_key()
     return delegating_priv, signing_priv
 
 
-@pytest.fixture(scope='function')
+@pytest.fixture
 def bobs_keys():
     priv = keys.UmbralPrivateKey.gen_key()
     pub = priv.get_pubkey()
-    return MockKeyPair(priv, pub)
+    return priv, pub
 
 
 @pytest.fixture()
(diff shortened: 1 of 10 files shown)


tests/conftest.py
```diff
@@ -90,24 +89,44 @@ def random_ec_curvebn2():
...
+
+
+@pytest.fixture(scope='session')
+def message():
+    message = b"dnunez [9:30 AM]" \
+              b"@Tux we had this super fruitful discussion last night with @jMyles @michwill @KPrasch" \
+              b"to sum up: the symmetric ciphertext is now called the 'Chimney'." \
+              b"the chimney of the capsule, of course" \
+              b"tux [9:32 AM]" \
+              b"wat"
```
michwill: Still loving that dialog!

tests/unit/test_umbral_keys.py
```diff
@@ -133,6 +119,18 @@ def test_key_encoder_decoder(random_ec_curvebn1):
     assert decoded_key.to_bytes() == umbral_key.to_bytes()
 
 
+def test_public_key_as_hex(random_ec_curvebn1):
+    pubkey = UmbralPrivateKey(random_ec_curvebn1, default_params()).get_pubkey()
+    hex_string = pubkey.hex()
```
michwill: Maybe also can add a check that it's indeed hex and nothing was messed up, such as:
```
assert set(hex_string) == set('0123456789abcdef')
assert len(hex_string) == 66
```

umbral/pre.py
```diff
@@ -46,6 +47,13 @@ def __init__(self, message: str, offending_cfrags: List[CapsuleFrag]) -> None:
         self.offending_cfrags = offending_cfrags
 
 
+class UmbralDecryptionError(GenericUmbralError):
+    def __init__(self) -> None:
+        super().__init__("Decryption of ciphertext failed: "
```
tuxxy: Not sure if this is really necessary here, can't you just do a `return` if you want this?

tests/unit/test_umbral_keys.py
cygnusv: ✅

umbral/pre.py
cygnusv: Not sure what you mean here 🤔

tests/unit/test_umbral_keys.py
```diff
@@ -163,11 +170,11 @@ def test_keying_material_serialization_with_encryption():
 
     insecure_cost = 15  # This is deliberately insecure, just to make the tests faster
     encoded_keying_material = umbral_keying_material.to_bytes(password=b'test',
-                                                  _scrypt_cost=insecure_cost)
+                                                              _scrypt_cost=insecure_cost)
```
KPrasch: Other than tests - is there any reason to provide `_scrypt_cost` as a kwarg?

tests/conftest.py
jMyles: :100:

tuxxy:
@cygnusv Wanna merge this?

PR #249, opened 2019-08-08
Facilities to reduce abuse of _attached_cfrags

jMyles:
Presently, there are several places in the NuCypher codebase (and on in this codebase) where we abuse the private `_attached_cfrags` attribute of `Capsule`.

This PR allows for the Capsule's CFrags to be cleared without needing to access this attribute.

### DIFF
--- a/umbral/pre.py
+++ b/umbral/pre.py
@@ -176,6 +176,9 @@ class Capsule:
             error_msg = "CFrag is not correct and cannot be attached to the Capsule"
             raise UmbralCorrectnessError(error_msg, [cfrag])
 
+    def clear_cfrags(self):
+        self._attached_cfrags = set()
+
     def first_cfrag(self):
         try:
             return list(self._attached_cfrags)[0]
--- a/umbral/pre.py
+++ b/umbral/pre.py
@@ -176,6 +176,12 @@ class Capsule:
             error_msg = "CFrag is not correct and cannot be attached to the Capsule"
             raise UmbralCorrectnessError(error_msg, [cfrag])
 
+    def first_cfrag(self):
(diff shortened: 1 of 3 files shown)


michwill:
Unrelated to the PR itself, but failing CI seems do be looking for `import sha3`?

umbral/pre.py
```diff
@@ -173,6 +173,9 @@ def components(self) -> Tuple[Point, Point, CurveBN]:
     def __bytes__(self) -> bytes:
         return self.to_bytes()
 
+    def __contains__(self, cfrag):
+        return cfrag in self._attached_cfrags
+
```
cygnusv: 😍

umbral/pre.py
```diff
@@ -378,7 +391,7 @@ def _decapsulate_reencrypted(receiving_privkey: UmbralPrivateKey, capsule: Capsu
     pub_key = receiving_privkey.get_pubkey().point_key
     priv_key = receiving_privkey.bn_key
 
-    precursor = capsule._attached_cfrags[0].point_precursor
+    precursor = capsule.first_cfrag().point_precursor
```
cygnusv: 🎉

umbral/pre.py
```diff
@@ -65,7 +66,7 @@ def __init__(self,
         self.point_v = point_v
         self.bn_sig = bn_sig
 
-        self._attached_cfrags = list()    # type: list
+        self._attached_cfrags = set()    # type: list
```
cygnusv:
```suggestion
        self._attached_cfrags = set()    # type: set
```
Just in case mypy complains later.

PR #251, opened 2019-10-17
add vectors and test for kdf function

vepkenez:


cygnusv:
Something unrelated but that makes `runt_tests_35` fail: the new release of `BytestringSplitter` uses f-strings, which are incompatible with python 3.5. We can either pin the previous version of BSS or just deprecate support to python 3.5, which we anyway don't support in `nucypher`

vepkenez:
Ah that's interesting I have only run any of this in 3.7 and everything passes.

tuxxy:
@cygnusv Did we ever deprecate Python 3.5 after this? The master build still shows as failing.

PR #261, opened 2021-01-21
Replace hazmat Blake2b with standard lib

vepkenez:
According to some of our potential app developers, the compilation of cryptography.hazmat.primitives.Blake2B is the sole obstacle to running NuCypher on ARM Cpus.

Maybe this will fix it.


fjarri:
Obsoleted by PR #263 - Switched to SHA2

PR #263, opened 2021-03-16
Interfacing with Rust

fjarri:
Refactor to achieve binary compatibility (and API uniformness) with `rust-umbral` (see the corresponding PR https://github.com/nucypher/rust-umbral/pull/41):

- drop support of Py 3.5
- remove all hashing algorithms but SHA256
- implement `unsafe_hash_to_point()` in the same way as `rust-umbral`
- implement `hash_to_curvebn()` in the same way RustCrypto does it
- use XChaCha instead of Salsa to store encrypted secret keys
- implement adding DST to hashes in the same way as `rust-umbral`
- use hashing with DST everywhere, with the same tags as `rust-umbral`
- require metadata on cfrag verification
- add compatibility tests checking cross-operation with Rust
- change API to make it a drop-in replacement for `rust-umbral`
- remove unused public API 
- add API documentation

**For reviewers:**

The size may seem daunting, but a lot of it is due to vector changes. Perhaps it will be better to review commit-by-commit, or just file by file. The size of both tests and the library itself shrunk by 30-40%, and 25% of the library is just OpenSSL wrapper functions.

I'm planning to make a test PR to `nucypher` to see if this new version has enough functionality for it. There may be slight changes in the API because of that, but the bulk of the changes is already made.

In particular, `BytestringSplitter` may have to be brought back. I removed it temporarily because I couldn't figure out splitter composition, and didn't want that to block me.

**Performance:**

There are two differences between this version and the main branch currently.
1. `generate_kfrags` is ~15% slower. This is caused by one secret key -> public key computation, which is cached in the main branch. Fixed, but in Python version only - just to avoid explaining why there's a performance regression. It starts getting less and less important the higher `m` and `n` you take.
2. `reencrypt` is ~2x faster. The main branch verifies the capsule twice (4 curve point * scalar multiplications), and kfrag once (1 curve point * scalar multiplication). This effectively doubles the number of multiplications, which are the main bottleneck in this operation. In this PR, capsules are verified on deserialization only, and `reencrypt()` does not attempt to verify kfrags (the caller must call verification explicitly).


docs/source/using_pyumbral.rst
```diff
@@ -12,30 +12,22 @@ Using pyUmbral
...
 
-.. testcleanup:: capsule_story
+Elliptic Curves
+===============
 
-    from umbral import config
-    config._CONFIG.___CONFIG__curve = None
-    config._CONFIG.___CONFIG__params = None
+The matter of which curve to use is the subject of some debate.  If you aren't sure, you might start here:
+https://safecurves.cr.yp.to/
```
derekpierre, line 19: 👍

docs/source/using_pyumbral.rst
```diff
@@ -137,22 +129,21 @@ or re-encrypted for him by Ursula, he will not be able to open it.
...
-    >>> fail = pre.decrypt(ciphertext=ciphertext,
-    ...                    capsule=capsule,
-    ...                    decrypting_key=bobs_private_key)
+    >>> fail = decrypt_original(sk=bobs_private_key,
+    ...                         capsule=capsule,
+    ...                         ciphertext=ciphertext)
     Traceback (most recent call last):
         ...
-    umbral.pre.UmbralDecryptionError
+    umbral.dem.ErrorInvalidTag
```
derekpierre, line 137: wdyt about `InvalidTagError` or just `InvalidTag` ? 😎

tests/test_capsule.py
```diff
@@ -0,0 +1,108 @@
...
+    with pytest.raises(ValueError, match="Empty CapsuleFrag sequence"):
+        capsule.open_reencrypted(receiving_sk, delegating_pk, [])
+
+    # Not enough cfrags
+    with pytest.raises(ValueError, match="Internal validation failed"):
+        capsule.open_reencrypted(receiving_sk, delegating_pk, cfrags[:threshold-1])
+
+    # Repeating cfrags
+    with pytest.raises(ValueError, match="Some of the CapsuleFrags are repeated"):
+        capsule.open_reencrypted(receiving_sk, delegating_pk, [cfrags[0]] + cfrags[:threshold-1])
```
derekpierre, line 92: 💥

derekpierre (commented):
Still working my way through but looks great so far!

vepkenez:
Looks excellent to me.  Amazing work!

docs/notebooks/pyUmbral Simple API.ipynb
```diff
@@ -184,21 +164,24 @@
...
-    "kfrags = pre.generate_kfrags(delegating_privkey=alices_private_key, \n",
-    "                             receiving_pubkey=bobs_public_key, \n",
-    "                             signer=alices_signer,\n",
-    "                             threshold=M, \n",
-    "                             N=N)\n"
+    "kfrags = generate_kfrags(delegating_sk=alices_private_key,\n",
+    "                         receiving_pk=bobs_public_key,\n",
+    "                         signing_sk=alices_signing_key,\n",
+    "                         threshold=10,\n",
+    "                         num_kfrags=20)"
```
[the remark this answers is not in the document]
fjarri: Used `M` and `N`, `M` is used later so I guess it makes sense.

docs/notebooks/pyUmbral Simple API.ipynb
```diff
@@ -236,15 +219,36 @@
...
+    "                        signing_pk=alices_verifying_key)\n",
+    "           for cfrag in cfrags)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Bob opens the capsule; Decrypts data from Alice.\n",
+    "Finally, Bob decrypts the re-encrypted ciphertext using his key."
```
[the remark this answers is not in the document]
fjarri: Fixed.

docs/source/index.rst
```diff
@@ -31,7 +31,7 @@ pyUmbral
 .. end-badges
 
 pyUmbral is the reference implementation of the Umbral_ threshold proxy re-encryption scheme.
-It is open-source, built with Python, and uses OpenSSL_ and Cryptography.io_.
+It is open-source, built with Python, and uses OpenSSL_ via Cryptography.io_ and libsodium_ via PyNaCl_.
```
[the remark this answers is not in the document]
fjarri: Fixed

docs/source/using_pyumbral.rst
```diff
@@ -12,30 +12,22 @@ Using pyUmbral
...
-The best way to start using pyUmbral is to decide on an elliptic curve to use and set it as your default.
-
-
-.. doctest:: capsule_story
+.. _Cryptography.io: https://cryptography.io/en/latest/
 
-    >>> from umbral import config
-    >>> from umbral.curve import SECP256K1
-    >>> config.set_default_curve(SECP256K1)
+Be careful when choosing a curve - the security of your application depends on it.
```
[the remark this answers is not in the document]
fjarri: Fixed.

docs/source/using_pyumbral.rst
```diff
@@ -93,29 +85,29 @@ Bob Exists
...
-    >>> bobs_public_key = bobs_private_key.get_pubkey()
+    >>> bobs_private_key = SecretKey.random()
+    >>> bobs_public_key = PublicKey.from_secret_key(bobs_private_key)
 
 
-Alice grants access to Bob by generating kfrags 
+Alice grants access to Bob by generating kfrags
 -----------------------------------------------
-When Alice wants to grant Bob access to open her encrypted messages, 
+When Alice wants to grant Bob access to open her encrypted messages,
```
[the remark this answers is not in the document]
fjarri: Fixed

docs/source/using_pyumbral.rst
fjarri: Yes, the what errors to raise is still something I'm thinking about. Should ideally be synchronized with rust-umbral too.

tests/test_capsule.py
fjarri: Better than getting a division by zero error somewhere in the depths of OpenSSL :)

docs/source/using_pyumbral.rst
fjarri: Now an `umbral.UmbralError` is raised.

umbral/keys.py
```diff
@@ -1,472 +1,205 @@
...
-
-
-def derive_key_from_password(password: bytes,
-                             salt: bytes,
-                             **kwargs) -> bytes:
+class SecretKey(Serializable):
     """
-    Derives a symmetric encryption key from a pair of password and salt.
-    It uses Scrypt by default.
+    Umbral secret (private) key.
```
cygnusv, line 21: Usually the term "private" key implicitly signals an asymmetric cryptography scheme (hinting at the private vs public dichotomy), while "secret" key is usually associated to symmetric cryptography.

umbral/openssl.py
```diff
@@ -1,218 +1,448 @@
...
+    bn = _bn_new(set_consttime_flag)
     backend._lib.BN_bin2bn(bytes_seq, len(bytes_seq), bn)
-    backend.openssl_assert(bn != backend._ffi.NULL)    
+    backend.openssl_assert(bn != backend._ffi.NULL)
+
+    if check_modulus and not bn_is_normalized(bn, check_modulus):
+        raise ValueError(f"The integer encoded with given bytes ({repr(bytes_seq)}) "
+                          "is not under the provided modulus.")
+
+    if apply_modulus:
```
cygnusv, line 180: This is an awesome idea 🤯

umbral/openssl.py
```diff
@@ -1,218 +1,448 @@
...
+        bignum =_bn_new()
+        with backend._tmp_bn_ctx() as bn_ctx:
+            res = backend._lib.BN_mod(bignum, bn, apply_modulus, bn_ctx)
+            backend.openssl_assert(res == 1)
```
cygnusv, line 184: It seems we're not doing anything with the `bignum` result later. It'd be cool that we could simply say `BN_mod(bn, bn, apply_modulus, bn_ctx)`, but it's not clear whether the API allows it.

umbral/keys.py
fjarri: I was following the name convention from `rust-umbral`, which, in turn, follows the terms used by the RustCrypto stack. Is using "secret" instead of "private" here very confusing, or just somewhat unusual?

As an extension of uniformity with RustCrypto, I was thinking of just re-exporting the RustCrypto key objects themselves instead of wrapping them in our own types. This way the keys can be used with other libraries based on RustCrypto, making it unnecessary to have conversion methods like `to_cryptography_privkey()` we have in PyUmbral.

Also being able to abbreviate to "sk" and "pk" is quite useful.

umbral/openssl.py
fjarri: Good catch, we should indeed return `bignum` here.

I don't think any of the OpenSSL functions support inplace operations, unfortunately (seeing as how bignums probably use dynamic allocations inside).

umbral/pre.py
```diff
@@ -1,518 +1,58 @@
...
-        # (i.e., this is an original capsule)
-        encapsulated_key = _decapsulate_original(decrypting_key, capsule)
-
-    dem = UmbralDEM(encapsulated_key)
-    try:
-        cleartext = dem.decrypt(ciphertext, authenticated_data=bytes(capsule))
-    except InvalidTag as e:
-        raise UmbralDecryptionError() from e
+    key_seed = capsule.open_reencrypted(decrypting_sk, delegating_pk, cfrags)
+    # TODO: add salt and info here?
```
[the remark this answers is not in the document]
fjarri: Fixed.

umbral/openssl.py
fjarri: Fixed.

Pipfile
```diff
@@ -5,13 +5,8 @@ name = "pypi"
 
 [packages]
 setuptools = "*"
-# Third Party
 cryptography = ">=2.3"
 pynacl = "*"
-pysha3 = "*"
-# NuCypher
-bytestring-splitter = "*"
-constant-sorrow = ">=0.1.0a7"
```
cygnusv, line 14: Yay!

tests/conftest.py
```diff
@@ -1,139 +1,51 @@
...
-              b"to sum up: the symmetric ciphertext is now called the 'Chimney'." \
-              b"the chimney of the capsule, of course" \
-              b"tux [9:32 AM]" \
-              b"wat"
+    message = (b"dnunez [9:30 AM]"
+               b"@Tux we had this super fruitful discussion last night with @jMyles @michwill @KPrasch"
+               b"to sum up: the symmetric ciphertext is now called the 'Chimney'."
+               b"the chimney of the capsule, of course"
+               b"tux [9:32 AM]"
+               b"wat")
```
cygnusv, line 37: good thing we're keeping this fixture

cygnusv (approved):
Thanks for the walkthrough yesterday! The PR looks great and to me it's a relief seeing this simplification in so many aspects of Umbral.

I think I gave you my main comments during the walkthrough and I see you are working on that in part 2. I still think the use of "SecretKey" is not the best choice taking into consideration the uses in the industry and academia for public-key cryptosystems. As a random example, look at the recent draft for Hybrid Public Key Encryption from the IETF (https://www.ietf.org/archive/id/draft-irtf-cfrg-hpke-08.html#name-notation); they use the term "private key" even though they use the `sk` abbreviation.

But that's something that's not critical at all in the scope of this PR.

fjarri:
> I think I gave you my main comments during the walkthrough and I see you are working on that in part 2. I still think the use of "SecretKey" is not the best choice taking into consideration the uses in the industry and academia for public-key cryptosystems.

Perhaps. Let us put a pin in that and revisit it after existing PRs — changing it in 4 PRs now will lead to horrible rebases.

I agree with following the existing conventions, but I wonder why RustCrypto chose to call it SecretKey - the author is usually very diligent in implementing crypto standards.

PR #267, opened 2021-04-19
Interfacing with Rust, part 2

fjarri:
More changes based on the needs of `nucypher`. Based on top of #263 

Synchronized with https://github.com/nucypher/rust-umbral/commit/3cba9022ff94f211bc5e1ad6d72f08282f8ff960

- Introducing Signer class and making `generate_kfrags()` take it instead of the secret key
- Some parameter name normalization (`signing_pk` -> `verifying_pk`)
- Removed all the supported curves other than secp256k1
- Removed DST for the hash used in signing/verification
- Moved `generate_kfrags()` to `pre.py` (where the rest of high-level API resides)
- Moved backend parts of signing to `openssl.py`
- `CurvePoint.from_affine()` is removed


umbral/signing.py
```diff
@@ -1,151 +1,143 @@
...
-
-        signature_der_bytes = self.__cryptography_private_key.sign(message, signature_algorithm)
-        return Signature.from_bytes(signature_der_bytes, der_encoded=True, curve=self.curve)
+    ge = CurvePoint.generator() * e
+    rs = point_r * s
+
+    # Try `r`
+    q = (rs - ge) * inv_r
+    if q == public_key.point():
+        return False
```
derekpierre, line 136: Noticed this yesterday during the walkthrough - how come this returns False?

fjarri: This function returns a recovery byte, which can only be `\x00` or `\x01`, so I return a boolean flag instead.

cygnusv (approved):

piotr-roslaniec (approved):

PR #270, opened 2021-06-02
Final touches for the API

fjarri:
Synced with https://github.com/nucypher/rust-umbral/pull/50

https://github.com/nucypher/nucypher/pull/2612 uses this branch in CI

- add `VerifiedKeyFrag.from_verified_bytes()` (for storage purposes)
- remove `metadata` support from `reencrypt()` (see https://github.com/nucypher/nucypher/issues/259)
- remove unused dependencies
- rename `pk` to `delegating_pk` in `encrypt()`
- rename `decrypting_sk` to `delegating_sk` in `decrypt_original()`
- rename `decrypting_sk` to `receiving_sk` in `decrypt_reencrypted()`
- expose serialized object sizes as `HasSerializedSize.serialized_size()`, and overhaul `Serializable` given the new assumption that all of our objects have constant sizes known in advance. The size is needed in `nucypher` (to set up splitters)
- in addition, split `Serializable` into `Serializable` and `Deserializable`
- replace `GenericError` with the standard `ValueError`

### DIFF
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -170,7 +170,7 @@ def _decrypt_reencrypted(umbral, receiving_sk_bytes, delegating_pk_bytes, verify
                        for cfrag in cfrags]
 
     # Decryption by Bob
-    plaintext = umbral.decrypt_reencrypted(decrypting_sk=receiving_sk,
+    plaintext = umbral.decrypt_reencrypted(receiving_sk=receiving_sk,
                                            delegating_pk=delegating_pk,
                                            capsule=capsule,
                                            verified_cfrags=verified_cfrags,
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -143,16 +143,16 @@ def test_kfrags(implementations):
 
 
 def _reencrypt(umbral, verifying_pk_bytes, delegating_pk_bytes, receiving_pk_bytes,
-               capsule_bytes, kfrags_bytes, threshold, metadata):
+               capsule_bytes, kfrags_bytes, threshold):
     capsule = umbral.Capsule.from_bytes(bytes(capsule_bytes))
     verified_kfrags = _verify_kfrags(umbral, kfrags_bytes,
                                      verifying_pk_bytes, delegating_pk_bytes, receiving_pk_bytes)
-    cfrags = [umbral.reencrypt(capsule, kfrag, metadata=metadata) for kfrag in verified_kfrags[:threshold]]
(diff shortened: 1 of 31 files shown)


tests/test_compatibility.py
```diff
@@ -243,3 +242,30 @@ def test_signer(implementations):
...
+        umbral.PublicKey,
+        umbral.Signature,
+        umbral.Capsule,
+        umbral.KeyFrag,
+        umbral.VerifiedKeyFrag,
+        umbral.CapsuleFrag,
+        umbral.VerifiedCapsuleFrag,
+        ]
+
+    return {tp.__name__: tp.serialized_size() for tp in sized_types}
```
vepkenez, line 261: Does this output some like:
```
{
   "SecretKey": 17
   ...
}
```?

fjarri: Yep. Can't just put a class there, because it will be different for the other implementation

cygnusv:
> - rename `pk` to `delegating_pk` in `encrypt()`

Rather than `delegating_pk`, this should be a simple `encrypting_pk`. One of the core propositions of PRE is that delegation is completely orthogonal to encryption.

> - rename `decrypting_sk` to `delegating_sk` in `decrypt_original()`
> - rename `decrypting_sk` to `receiving_sk` in `decrypt_reencrypted()`

I don't get this distinction between delegating and receiving. In both cases it's "receiving"; "decrypting" also works.

fjarri:
> I don't get this distinction between delegating and receiving. In both cases it's "receiving"; "decrypting" also works.

Here's my rationale: 

First, the `encrypting/decrypting` prefixes. I think they are redundant; if you are passing a key to an `encrypt/decrypt`, these keys are by default used for encrypting/decrypting (while other usages still warrant a prefix, like `verifying_key`).

Now for the `delegating/receiving` prefixes. While it's true that one can use Umbral as a simple asymmetric encryption scheme (and we do use it that way to encrypt kfrags for Ursula), it is not the intended usage, and, in fact, an overkill. I wanted to emphasize that:
- during normal operation `encrypt` is supposed to take a key the delegator, and not the receiver, has control over (I was confused by this myself on several occasions)
- `decrypt_original` is more of a debugging tool than a part of a normal process, and takes the same key that was used in `encrypt`
- `decrypt_reencrypted` takes a different key compared to what was used in `encrypt`

piotr-roslaniec (approved):

cygnusv (approved):

PR #271, opened 2021-06-13
Add `SecretKey.public_key()` shortcut

fjarri:
Paired with https://github.com/nucypher/rust-umbral/pull/52

This is technically a minor change, but it has a pretty big surface.

This PR removes `PublicKey.from_secret_key()` constructor and replaces it with `SecretKey.public_key()` shortcut. Bundled with it are some minor type hint fixes and a dependency bump.

So, why  the change? Creating a public key out of a secret key is a very common operation, both in `PyUmbral` itself, and in `nucypher`. Typing `PublicKey.from_secret_key()` can get repetitive, and you cannot chain it easily (with `SecretKey.random()`, for example). So a deviation from the "every type is responsible for creating its objects" guideline is justified here. Kind of like a [desire path](https://en.wikipedia.org/wiki/Desire_path) — no reason to fight it. Following that, if we already have a way to create a public key from a secret key, there is no need for a second way (`PublicKey.from_secret_key()`), so it is being removed. 

Incidentally, `RustCrypto` [uses the same API](https://docs.rs/elliptic-curve/0.10.1/elliptic_curve/struct.SecretKey.html#method.public_key), which gives additional support to this change.

Additionally, `cryptography` is being restricted to `3.*`, `pynacl` to `1.*` and `sphinx` to `4.*`. The latter finally fixes the problem with conflicts between the typehints that Sphinx's autodoc understands, and the typehints that `mypy` understands.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,8 +5,8 @@ name = "pypi"
 
 [packages]
 setuptools = "*"
-cryptography = ">=2.3"
-pynacl = "*"
+cryptography = "~=3.0"
+pynacl = "~=1.0"
 
 [dev-packages]
 bumpversion = "*"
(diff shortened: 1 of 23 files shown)


piotr-roslaniec (approved):

derekpierre (approved):
🎸 - Much better flow for `secet_key.public_key()` than `PublicKey.from_secret_key()`.

PR #272, opened 2021-07-19
Api updates corresponding to post-0.2 `rust-umbral` PRs

fjarri:
Already merged `rust-umbral` PRs:
- Made `SecretKey` and `SecretKeyFactory` serializable via `to_secret_bytes()` instead of `bytes()`, and added a `SerializableSecret` base class (https://github.com/nucypher/rust-umbral/pull/53)
- Removed `SecretKey.__eq__()` (https://github.com/nucypher/rust-umbral/pull/53)
- Added `VerifiedCapsuleFrag.from_verified_bytes()` (https://github.com/nucypher/rust-umbral/pull/63)
- Changed the seed size in SecretKeyFactory from 64 to 32 bytes (https://github.com/nucypher/rust-umbral/pull/64)
- Added `SecretKeyFactory.secret_key_factory_by_label()` (https://github.com/nucypher/rust-umbral/pull/64)
- Added `SecretKeyFactory.from_secure_randomness()` and `.seed_size()` (https://github.com/nucypher/rust-umbral/pull/64)

### DIFF
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -22,7 +22,7 @@ def pytest_generate_tests(metafunc):
 def _create_keypair(umbral):
     sk = umbral.SecretKey.random()
     pk = sk.public_key()
-    return bytes(sk), bytes(pk)
+    return sk.to_secret_bytes(), bytes(pk)
 
 
 def _restore_keys(umbral, sk_bytes, pk_bytes):
@@ -45,7 +45,7 @@ def test_keys(implementations):
 def _create_sk_factory_and_sk(umbral, label):
     skf = umbral.SecretKeyFactory.random()
     sk = skf.secret_key_by_label(label)
-    return bytes(skf), bytes(sk)
+    return skf.to_secret_bytes(), sk.to_secret_bytes()
 
 
 def _check_sk_is_same(umbral, label, skf_bytes, sk_bytes):
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -42,25 +42,32 @@ def test_keys(implementations):
     _restore_keys(umbral2, sk_bytes, pk_bytes)
 
 
-def _create_sk_factory_and_sk(umbral, label):
+def _create_sk_factory_and_sk(umbral, skf_label, key_label):
     skf = umbral.SecretKeyFactory.random()
-    sk = skf.secret_key_by_label(label)
-    return skf.to_secret_bytes(), sk.to_secret_bytes()
+    derived_skf = skf.secret_key_factory_by_label(skf_label)
+    sk = derived_skf.secret_key_by_label(key_label)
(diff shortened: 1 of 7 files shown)


tests/test_compatibility.py
```diff
@@ -42,25 +42,32 @@ def test_keys(implementations):
     _restore_keys(umbral2, sk_bytes, pk_bytes)
 
 
-def _create_sk_factory_and_sk(umbral, label):
+def _create_sk_factory_and_sk(umbral, skf_label, key_label):
     skf = umbral.SecretKeyFactory.random()
-    sk = skf.secret_key_by_label(label)
-    return bytes(skf), bytes(sk)
+    derived_skf = skf.secret_key_factory_by_label(skf_label)
+    sk = derived_skf.secret_key_by_label(key_label)
```
cygnusv, line 48: Should it be `by_label` or `from_label`?

umbral/keys.py
```diff
@@ -53,7 +50,7 @@ def serialized_size(cls):
     def _from_exact_bytes(cls, data: bytes):
         return cls(CurveScalar._from_exact_bytes(data))
 
-    def __bytes__(self) -> bytes:
+    def to_secret_bytes(self) -> bytes:
```
cygnusv, line 53: 👍

umbral/keys.py
```diff
@@ -125,6 +145,14 @@ def secret_key_by_label(self, label: bytes) -> SecretKey:
```
cygnusv, line 145: Not related to this OR but we discussed this extra hoop before in discord (https://discord.com/channels/411401661714792449/411401661714792451/860193967923527730)

tests/test_compatibility.py
piotr-roslaniec: That would also be my guess, but I think `from` prefix is widely accepted as a naming convention for [`From<>` trait implementation](https://ricardomartins.cc/2016/08/03/convenient_and_idiomatic_conversions_in_rust).

piotr-roslaniec (approved):

tests/test_compatibility.py
fjarri: "by_label" is a leftover from PyUmbral 0.1. `from` is indeed used generally for constructors.

Come to think of it, perhaps `make_secret_key()`/`make_secret_key_factory()` (or `derive_`, borrowing from PyUmbral 0.1) would be a better name for these methods. `by_label` in my mind implies that there is some lookup going on and not creation.

umbral/keys.py
fjarri: Yep, I remember - see my concerns in https://github.com/nucypher/rust-umbral/pull/64

fjarri:
The discussion about hashing to scalars is happening in https://github.com/nucypher/rust-umbral/issues/35 . The rest is ok to mere since https://github.com/nucypher/rust-umbral/pull/64 is merged.

